Issue #1, opened 2023-11-20
Simulation setup steps checklist


TomNicholas:
We want to translate all the MATLAB setup tools into Python, including:

- [x] Step 1: Make grid
- [ ] Step 2: Make topography
- [ ] Step 3: Make tides
- [ ] Step 4: Make atmospheric forcing
- [ ] Step 5: Make initial conditions
- [ ] Step 6: Make boundary forcing
- [ ] Step 7: Set up nesting (optional for user)

PR #4, opened 2023-11-28
Add topography

TomNicholas:
Towards step (2) of issue #1.

Currently a work-in-progress, so the tests will fail right now.

I want to avoid packaging the `etopo5.nc` bathymetry file as part of this repository, so instead I propose we use a package called [`pooch`](https://github.com/fatiando/pooch) to automatically "fetch" and download this file from somewhere else on the internet (which could just be a second separate, dedicated repository for storing files) only as and when the file is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Check the installation has worked by running the tests (you will need to also in
 pytest
 ```
 
-Dependencies required are xarray and netcdf4, plus matplotlib and cartopy for visualising grids.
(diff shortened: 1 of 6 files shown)


TomNicholas:
`etopo5.nc` now lives in the new repository [`roms-tools-data`](https://github.com/CWorthy-ocean/roms-tools-data)

TomNicholas:
Merging this now so Dafydd can use it

PR #12, opened 2024-05-03
Update conda environment and README

NoraLoose:

PR #14, opened 2024-05-06
Cleanup grid generation

NoraLoose:
* Fixes typo in grid dataset generation
* Gets rid off dummy dimension "one"

* Also adds some more packages to the conda enviroment (see https://github.com/CWorthy-ocean/roms-tools/pull/12)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ _Note these tools are for the [Center for Earth Systems Research Group](http://r
 pip install roms-tools
 ```
 
-### Installation from GitHub
+### Install from GitHub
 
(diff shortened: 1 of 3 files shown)


NoraLoose:
Tests fail because conda environment called in https://github.com/CWorthy-ocean/roms-tools/blob/main/.github/workflows/tests.yaml (aka https://github.com/CWorthy-ocean/roms-tools/blob/main/ci/environment.yml) does not have all required dependencies. This PR should fix this issue.

NoraLoose:
I'm merging this because it is a minor fix.

Issue #9, opened 2024-01-17
What discrete choices get made during setup?


TomNicholas:
The matlab tools have 6 separate steps, but that doesn't mean that the ROMS user actually needs to make a new choice at each of those steps. For example when you create the grid (step 1) you choose the location on the earth of the domain, but when you make the topography (step 2) you don't make a new choice, the topography just has to be for the region you chose in step 1.

We should try to fuse together different steps in this workflow so that the user is only required to do anything when there is an actual new choice to be made.

How many discrete choices are there? At least:
1) Choose the resolution, size, and location of the domain,
2) Choose the time resolution you want to run with (needed for tides + forcing data?),
3) Choose which data sources you want to use for the initial/boundary/forcing data

Are there any others? What about choice of processor decomposition?

@ubbu36

NoraLoose:
> For example when you create the grid (step 1) you choose the location on the earth of the domain, but when you make the topography (step 2) you don't make a new choice, the topography just has to be for the region you chose in step 1.

There is actually a new decision involved when adding the topography, namely which data source to use. The matlab code allows 8 different choices, see https://github.com/CWorthy-ocean/C-Star/blob/6dfebe86eba059f3732764fa8ff99d9cd412290a/setup_tools/setup_r2r_phys%2Bbgc/2.Make_topo/topo.m#L57-L80.

 @sdbachman @TomNicholas  Do we want to give the user the option to choose from different data sources for the topography? Probably yes... If so, we need to 
* upload more files to https://github.com/CWorthy-ocean/roms-tools-data. (Currently, we only have one, namely `etopo5.nc`.) Who can provide these files?
* modify the grid object to take one more parameter, something like
```
grid = Grid(
    nx=100,          # number of points in the x-direction (not including 2 boundary cells on either end)
    ny=100,          # number of points in the y-direction (not including 2 boundary cells on either end)
    size_x=1800,     # size of the domain in the x-direction (in km)
    size_y=2400,     # size of the domain in the y-direction (in km)
    center_lon=-21,  # longitude of the center of the domain
    center_lat=61,   # latitude of the center of the domain
    rot=20,          # rotation of the grid's x-direction from lines of constant longitude, with positive values being a counter-clockwise rotation
    topography_source = "etopo5" 
)
```

TomNicholas:
I feel like we discussed this a while ago and thought that in practice topography is unlikely to be something that anyone really cares about changing...

TomNicholas:
[...]

PR #27, opened 2024-05-14
Making topography

NoraLoose:
Completes Step 2 in #1 via the following four sub-steps:
1. Interpolate from specified topography file (`_make_raw_topography`)
2. Smooth topography globally to avoid grid-scale instabilities (`_smooth_topography_globally`)
3. Fill enclosed basins with land (`_fill_enclosed_basins`)
4. Smooth topography locally to satisfy `r < rmax` (` _smooth_topography_locally`)

Steps 1, 3, and 4 closely follow the matlab routines. For step 2, we deviate from the matlab implementation and instead use the package [GCM-Filters](https://gcm-filters.readthedocs.io/en/latest/).

The four steps for an example domain:

Step1
Step2
Step3
Step4

### Saved fields

Following the matlab routines, we save two fields (added as xarray variables to `grid.ds`):
* `hraw`: topography after step 2
* `h`: topography after step 4

### Enhancement of `Grid` object and default options

As we discussed in #9, the topography (with its four sub-steps) is made as part of the `Grid` object. The `Grid` object now takes a few extra kwargs:
```

topography_source : str, optional
        Specifies the data source to use for the topography. Options are "etopo5". The default is "etopo5".
smooth_factor: int
        The smoothing factor used in the global Gaussian smoothing of the topography. The default is 2.
hmin: float
        The minimum ocean depth (in meters). The default is 5.
rmax: float
        The maximum slope parameter (in meters). The default is 0.2.
```

###  Smooth factor: python vs. matlab
Choosing `smooth_factor` = 2 as a default in the python implementation may seem relatively small compared to the matlab default of choosing a smoothing factor / width of 8 (or radius of 4). However, upon comparing the python- and matlab-generated final topographies with their respective default smoothing factors, the python-generated topography is still smoother than the matlab-generated one:
Screenshot 2024-05-14 at 12 27 16 PM

This is due to different smoother implementations. The python implementation uses a smoother with a Gaussian kernel has the same standard deviation as a boxcar filter with a kernel width (not half-width) equal to `smooth_factor` grid points (see the [GCM-Filters convention](https://gcm-filters.readthedocs.io/en/latest/theory.html#filter-scale-and-shape) or Section 2.3.4 in [Grooms et al. (2021)](https://onlinelibrary.wiley.com/doi/abs/10.1029/2021MS002552)), while the matlab implementation does something different. We could consider setting the `smooth_factor` default to something lower than 2 to get closer to the matlab result, but the GCM-Filter filters get less and less accurate if we approach `smooth_factor = 1`. To be discussed!


NoraLoose:
@TomNicholas @sdbachman if you have feedback, please let me know!

NoraLoose:
@sdbachman looked over this and gave me a thumbs-up. Scott, I may add some progress reports as part of a later PR, but I am merging this for now.

PR #32, opened 2024-05-21
Tidal forcing

NoraLoose:
This PR completes step 3 in #1.

The tidal forcing depends on the grid, and can be created via
```
from roms_tools import Grid, TidalForcing
grid = Grid(
    nx=100,                   
    ny=100,                  
    size_x=1800,          
    size_y=2400,          
    center_lon=0,          
    center_lat=61,        
    rot=20,                 
)
tidal_forcing = TidalForcing(grid)
tidal_forcing.plot("ssh_Re", nc=0)
```
Screenshot 2024-05-21 at 8 46 24 AM

### Still to do

The TPXO data is currently read in locally from the NCAR machines. We need to put the TPXO data in an AWS bucket in the cloud, and fetch it from there.

### Comparison to matlab-generated tidal forcing

The python-generated tidal forcing reproduces the matlab-generated tidal forcing up to differences that are many orders of magnitudes smaller than the signal, due to the use of different interpolation schemes for interpolating the data from the TPXO atlans:

Screenshot 2024-05-21 at 8 44 20 AM
Screenshot 2024-05-21 at 8 44 05 AM

The only significant differences are found for the tidal components 8, 9, 10, 11, 12 (the long-term tides) due to a bug in the *matlab* scripts, which was reported to Jeroen.


NoraLoose:
@TomNicholas skimmed through this PR, and gave me a 👍 to merge.

PR #35, opened 2024-06-10
Grid straddle and generating topography separately

NoraLoose:
This PR adds two features:
* The `grid.straddle` attribute that indicates whether or not the Greenwhich meridian crosses the domain. This attribute is helpful for interpolation tasks later on.
* The user has now the option to read in a grid from file, but use roms-tools to add the topography. Before this PR, grid generation and topography generation were handled as a single step.


NoraLoose:
These are relatively minor changes to the code. I am merging...

PR #54, opened 2024-07-11
Initial conditions from GLORYS

NoraLoose:
This PR 
* adds the `InitialConditions` object with interpolation, plotting, and saving capabilities
* adds some documentation on how to use it
* unifies some of the code in `initial_conditions.py` and `atmospheric_forcing.py` by using shared classes.

### Python vs. MATLAB-generated initial conditions

We follow the same strategy as the MATLAB tools, namely extrapolating the deepest value all the way down to the bottom, before doing the interpolation. This prevents local overturning (which could be introduced if we diffused in the horizontal to eliminate NaNs.)

The differences of the Python vs. MATLAB-generated initial conditions are relatively small, likely due to different interpolation methods. Here are a few examples:

Screenshot 2024-07-11 at 11 34 06 AM
Screenshot 2024-07-11 at 11 34 27 AM
Screenshot 2024-07-11 at 11 34 43 AM
Screenshot 2024-07-11 at 11 35 01 AM

Differences are perhaps the largest along the continental boundaries in the velocity fields:

Screenshot 2024-07-11 at 11 36 38 AM

Whereas differences for the tracer fields are pretty small:

Screenshot 2024-07-11 at 11 40 20 AM


NoraLoose:
I am merging this so I can work on some automated testing using the refactored code from this PR.

Issue #77, opened 2024-08-13
`spherical` missing from Grid


dafyddstephenson:
In the matlab-generated grid files we have:
```
	char spherical(one) ;
		spherical:Long_name = "Grid type logical switch" ;
		spherical:option_T = "spherical" ;
```
without which ROMS will not run:
```
 ### ERROR: get_grid :: Cannot find variable 'spherical' within netCDF file 'INPUT/PARTED/roms_grd.0.nc'.
```

NoraLoose:
This issue is solved.

Issue #71, opened 2024-08-13
Combine `Grid` and `VerticalCoordinate` objects


NoraLoose:
@dafyddstephenson and I just discussed to potentially combine the `Grid` and `VerticalCoordinate` objects into one object. The `.save()` method of the combined object would then write a netcdf file that has the following coordinates:
* lat/lon information at tracer, u- and v-points
* layer and interface depth at tracer, u- and v-points

While ROMS does not need any of these information (except lat information to compute the Coriolis parameter), this will be a useful reference file for the user to make plots etc.

As of now these coordinates are written as part of the initial conditions, boundary forcing files, etc. It would be better to have them all in one place, and maybe not clutter the ROMS input files with (for the model) unnecessary information.

dafyddstephenson:
I would float that we extend this to the boundary variables that aren't directly needed by the model too (`xi_rho_south` etc.), such that the grid file becomes a single reference file from which the user can get _all_ grid-related information for analysis.

NoraLoose:
Closing this because addressed in #81. I still have to remove the coordinates from the initial conditions, boundary forcing, and surface forcing input files, but this issue is tracked in #68.

PR #81, opened 2024-08-21
Updates to `Grid` object

### LINKED ISSUE #72: Apply `grid.coarsen()` automatically
The current workflow is interrupted by the need for the user to go back and apply `grid.coarsen()`, at step 5 (if they wish to use coarse forcing) modifying the grid object that they have already created by adding additional variables related to coarsening. If there isn't too much overhead, I believe this should be done automatically at the point where the grid object is made. Perhaps if the user is really trying to trim down their space requirements we can include an option not to do this, but I think the default behaviour should be to coarsen without user involvement.

NoraLoose:
This PR
* combines the `Grid` and `VerticalCoordinate` objects into one (see #71; a future PR will remove coordinates from initial conditions files etc.)
* removes `smooth_factor` and `rmax` options for the user since these parameters should not really be changed
* removes the method that computes the bathymetry Laplacian because it seems unnecessary as a result of the previous bullet point 
* closes #72 
* addresses #77 by adding the `spherical` variable to the grid. Hopefully ROMS will not complain about the fact that I did not give `spherical` a dimension `one` but simply save it as a scalar
* updates the documentation and method section
* adds more automated tests

cc @dafyddstephenson


NoraLoose:
I'm merging this so that @dafyddstephenson can test how the updated grid file format plays with ROMS.

PR #101, opened 2024-08-28
Tool for partitioning files

### LINKED ISSUE #97: Make partitioning tool for input files
## Issue

ROMS requires partitioned (or tiled) input files. The ROMS-Tools generated files are currently not partitioned but global. One option to get from global ROMS-Tools generated to partitioned files is to use the fortran script `partit.F`. However, it seems that the fortran `partit.F` does not play well with the ROMS-Tools generated files, see #92, #93, #95.

## Plan

It would there be nice to write a python / xarray version of `partit.F` **within ROMS-Tools**, which works seamlessly for both
- [ ] MATLAB-generated input files
- [ ] ROMS-Tools generated input files

Besides partitioning already created NetCDF input files (see above), it would be nice if the python version of `partit.F` could be called (optionally) as part the `.save()` method of the ROMS-Tools objects. I imagine the following:
```
InitialConditions.save(filepath)
```
will save a global initial conditions file per default - as currently done. However
```
InitialConditions.save(filepath=filepath, tiles={"nx": 3, "ny": 3})
```
will save a partitioned initial conditions file.

TomNicholas:
Will leave the ncjoin problem for a follow-up PR.

I've only tested this on a grid dataset created with roms-tools so far. It remains to be seen if there are differences in other types of input files which might need to be taken account of in this partitioning tool.

- [x] Closes #97
- [x] Tests
- [x] Passes pre-commit
- [x] Exposed as public API
- [ ] Documentation


roms_tools/utils.py
```diff
@@ -0,0 +1,125 @@
+from numbers import Integral
+
+import numpy as np
+import xarray as xr
+
+
+def partition(ds: xr.Dataset, nx: int = 1, ny: int = 1) -> list[xr.Dataset]:
+    """
+    Split a ROMS dataset up into nx by ny spatial tiles.
+    """
```
TomNicholas, line 10: We should expose a partitioning tool in the public API (as part of a new `utils` module perhaps), but I'm not sure if it should operate on files directly or on xarray Datasets (like this function does).

TomNicholas:
I'm not sure what the best way to test this is. One neat way would be to also write the `ncjoin` replacement, then assert that splitting then joining gives the same result as you started with.

NoraLoose:
Thanks for working on this @TomNicholas! It looks great - I will leave some more specific comments.

> I'm not sure what the best way to test this is. One neat way would be to also write the ncjoin replacement, then assert that splitting then joining gives the same result as you started with.

That would be the ultimate test. Until then we could just write a regression test, i.e., check that the partitioned data equals the expected partitioned data.

roms_tools/utils.py
NoraLoose: I think it's great to have it (more generally) work on xarray Datasets. Then we can invoke it as part of the `.save` method. 

But we should definitiely add a page to the documentation that illustrates how you would use `partition` on already saved files, via
```
xr.open_dataset()
partition()
xr.save_mfdataset()
```

roms_tools/utils.py
```diff
@@ -0,0 +1,125 @@
...
+            raise ValueError(
+                f"Partitioning nx = {nx} ny = {ny} does not divide the domain into subdomains of integer size."
+            )
+
+    eta_rho_domain_size = integer_division_or_raise(
+        ds.sizes["eta_rho"] - 2 * n_eta_ghost_cells, nx
+    )
+    eta_v_domain_size = integer_division_or_raise(
+        ds.sizes["eta_v"] - 1 * n_eta_ghost_cells, nx
+    )
```
NoraLoose, line 41: Not all ROMS input datasets will have the `eta_v` dimension. In fact, the only dimensions that will always be there are `eta_rho` and `xi_rho`. So we should make the partitioned sizes conditioned on whatever dimensions are present.

NoraLoose:
Should we invoke `partition` within the `.save` method of all the `ROMS-Tools` objects (`Grid`, `TidalForcing`, `SurfaceForcing`, `InitialConditions`, `BoundaryForcing`), or do you want this to be part of a future PR?

TomNicholas:
> Should we invoke partition within the .save method of all the ROMS-Tools objects (Grid, TidalForcing, SurfaceForcing, InitialConditions, BoundaryForcing), or do you want this to be part of a future PR?

Let's leave that for another PR, getting this function right is worth a PR on its own.

> That would be the ultimate test. Until then we could just write a regression test, i.e., check that the partitioned data equals the expected partitioned data.

Defining all that expected data in the tests would be annoying, so for now I've just tested that the sizes are as expected.

Notice I've also moved all `setup`-related tests into their own subdirectory (because I wanted to call the tests file for this `test_utils.py`, but that was already taken for testing `setup`-specific utils.

TomNicholas:
I don't understand why this is happening in the CI

```
============================= test session starts ==============================
platform linux -- Python 3.10.14, pytest-8.3.2, pluggy-1.5.0 -- /home/runner/micromamba/envs/romstools/bin/python
cachedir: .pytest_cache
rootdir: /home/runner/work/roms-tools/roms-tools
configfile: pyproject.toml
ERROR: file or directory not found: roms_tools/tests/test_datasets.py
collecting ... collected 0 items

============================ no tests ran in 0.00s =============================

/home/runner/micromamba/envs/romstools/lib/python3.10/site-packages/coverage/control.py:894: CoverageWarning: No data was collected. (no-data-collected)
  self._warn("No data was collected.", slug="no-data-collected")
Error: Process completed with exit code 4.
```

roms_tools/utils.py
TomNicholas: Addressed in https://github.com/CWorthy-ocean/roms-tools/pull/101/commits/981cb46981d94e57d8bb1a0504b086b9e95ed48c

roms_tools/utils.py
```diff
@@ -0,0 +1,169 @@
...
+                    int(xi_coarse_partition_indices[j]),
+                    int(xi_coarse_partition_indices[j + 1]),
+                )
+
+            partitioned_ds = ds.isel(**indexers)
+
+            partitioned_datasets.append(partitioned_ds)
+
+    # TODO return the file number index too?
+    return partitioned_datasets
```
TomNicholas, line 169: The reason we might want to return the file number index is that `xr.save_mfdataset` requires a list of file names as well as a list of datasets.

TomNicholas: Now returns a list of file numbers too, see https://github.com/CWorthy-ocean/roms-tools/pull/101/commits/315cd3ef91eb36f158f9c4e702df73b25216df68

TomNicholas:
I think this is ready to merge, at least as something to build upon.

roms_tools/utils.py
```diff
@@ -0,0 +1,172 @@
...
+) -> tuple[list[int], list[xr.Dataset]]:
+    """
+    Split a ROMS dataset up into nx by ny spatial tiles.
+    """
+
+    # TODO also check they are positive integers
+    if not isinstance(nx, Integral) or not isinstance(ny, Integral):
+        raise ValueError("nx and ny must be integers")
+
+    # 'eta_rho' and 'xi_rho' are always expected to be present
```
NoraLoose, line 18: I just remembered that Ulla was trying to partition her river input file but couldn't do it because that file is missing `eta_rho` and `xi_rho` (see discussion on slack today, https://cworthyworkspace.slack.com/archives/C06M8FCAYPN/p1724925409666779).

While ROMS-Tools is currently not producing a river input file (or anything without the dimensions `eta_rho` and `xi_rho`), we maybe want `partition` to work on more general files. In the extreme case, if none of the spatial dimensions are present, the dataset will just be copied x number of times, where x is the number of tiles.

roms_tools/utils.py
```diff
@@ -0,0 +1,172 @@
...
+        n_xi_ghost_cells = 0
+    else:
+        n_xi_ghost_cells = 1
+
+    def integer_division_or_raise(a: int, b: int) -> int:
+        remainder = a % b
+        if remainder == 0:
+            return a // b
+        else:
+            raise ValueError(
```
NoraLoose, line 42: Maybe it's worth including a test that checks that this ValueError is raised.

NoraLoose:
Looks good! I just left two more comments.

roms_tools/utils.py
TomNicholas: Done in https://github.com/CWorthy-ocean/roms-tools/pull/101/commits/50d1422284a995a92cbe2f0515ca4fdf7203a641

roms_tools/utils.py
TomNicholas: Done in https://github.com/CWorthy-ocean/roms-tools/pull/101/commits/f57788020d7539a3367bf70d8259e9ca9ea6860d

NoraLoose:
I'm happy to merge, but looks like one of the tests is stuck. I'm already working on refactoring the tests and making them faster - sorry about that!

PR #106, opened 2024-08-29
Stretching curves

NoraLoose:
## ROMS-internal Issue

The stretching curves `Cs_r`, `Cs_w` form the base of ROMS' vertical coordinate system. ROMS computes the stretching curves internally, based on the provided parameters `theta_s`, `theta_b`, `Tcline` in `roms.in`. To make sure that the user used the same vertical coordinate system to prepare their input data, ROMS checks against provided `Cs_r`, `Cs_w` values via a call of `checkdims.F` in the following two instances:
1) the grid file: https://github.com/CESR-lab/ucla-roms/blob/2a4e9107ffc771d2833fbea3e08f1e169759b5ff/src/grid.F#L429
2) the initial conditions file: https://github.com/CESR-lab/ucla-roms/blob/2a4e9107ffc771d2833fbea3e08f1e169759b5ff/src/get_init.F#L77

Even though the grid file vs. initial conditions file contain the exact same values for `Cs_r` (similarly for `Cs_w`), the check fails in 1), but passes in 2). The provided values for `Cs_r` match the ROMS-internal values up to 1e-8 (if given at single precision) and up to 1e-16 (if given at double precision), similar for `Cs_w`. The permitted difference in ROMS is given by `epsil`, which is set to 1e-7, so there should not be a problem in either case.

Long story short, I think there is a bug in the ROMS source code when calling `checkdims.F` from `grid.F`. Note that the matlab-tools only write `Cs_r` and `Cs_w` into the initial conditions file (but not into the grid file), so that bug may have never been detected before.

## ROMS-Tools workaround

For completeness, we still want to have the `Cs_r` and `Cs_w` values in the grid file. But to avoid the check (which would fail), I am renaming them to `CS_r` and `CS_w`. This is a little ugly, and down the road we will want to fix the bug in the ROMS source code. I am still writing the `Cs_r` and `Cs_w` values into the initial conditions file under their correct name ("`Cs_r`" and "`Cs_w`"), so that ROMS will actually perform the sanity check.


NoraLoose:
cc @dafyddstephenson @TomNicholas: This PR is one example for how we have to find work-arounds within ROMS-Tools if we want to avoid fixing bugs in the ROMS source code (for now).

dafyddstephenson:
I'd leave this open for now as hopefully we can get [this](https://github.com/CESR-lab/ucla-roms/pull/37) patched in quickly in ROMS.

NoraLoose:
Can we merge this @dafyddstephenson? The convention is now that `Cs_r` and `Cs_w` are written both into the grid file and the initial conditions file. That means ROMS-Tools generated grid files will only be compatible with the newest version of ROMS. Is that okay?

PR #109, opened 2024-09-03
Integrating `partition` and `.save()`

NoraLoose:
This PR is a follow-up of #101. It
* fixes a small bug that made `partition` return the wrong `file_numbers`
* integrates `partition` with the `.save()` methods
* adds some examples of how to use `partition` to the documentation


TomNicholas (approved):
Nice!

Issue #112, opened 2024-09-04
test for bit-for-bit reproduction with to/from yaml


dafyddstephenson:
I am still unable to get a bit-for-bit reproduction of tides, surface forcing, initial and boundary conditions by saving to netcdf and yaml, then loading from yaml, saving to netcdf again and comparing the checksums of the two netcdf files (works for the grid though!)

We should aim to test for this.

NoraLoose:
This bit-by-bit reproducibility test was implemented before we merged #140, but I had to comment it out for reasons explained in #142. I think it is okay to close this issue because we have #142 open. Agreed @dafyddstephenson?

PR #115, opened 2024-09-05
Check for bit-by-bit reproduction via file hashes

NoraLoose:
Addresses #112.

@dafyddstephenson I added tests for all `Objects` that compare the file hashes of `filename1` and `filename2`, where
```
object1.save(filename1)
object1.to_yaml(yaml_filename)
object2 = Object.from_yaml(yaml_filename)
object2.save(filename2)
```
All the checks pass, see below. Can you post an example for when the hashes do not coincide?


NoraLoose:
I'm merging this. @dafyddstephenson If you encounter further problems to reproduce files with identical hashes, please report in #112.

Issue #61, opened 2024-07-19
Pytest fails in multi-threaded context


NoraLoose:
When I run 
```
pytest test_atmospheric_forcing.py
```
on perlmutter I run into the following error:
```
Thread 0x00007f8ff1163700 (most recent call first):
  File "/global/homes/n/nloose/.conda/envs/romstools/lib/python3.12/concurrent/futures/thread.py", line 89 in _worker
  File "/global/homes/n/nloose/.conda/envs/romstools/lib/python3.12/threading.py", line 1010 in run
  File "/global/homes/n/nloose/.conda/envs/romstools/lib/python3.12/threading.py", line 1073 in _bootstrap_inner
  File "/global/homes/n/nloose/.conda/envs/romstools/lib/python3.12/threading.py", line 1030 in _bootstrap
```

In contrast, if I execute all tests within `test_atmospheric_forcing.py` sequentially, i.e.,
```
pytest test_atmospheric_forcing.py::test_successful_initialization_with_regional_data
pytest test_atmospheric_forcing.py::test_no_longitude_intersection_initialization_with_regional_data
...
```
the entire test suite runs successfully.

What's going on?

NoraLoose:
This happens when running the test suite on perlmutter. Need to be fixed before the C-Star alpha release. Any ideas for what's going on @TomNicholas?

TomNicholas:
I don't know why that happens, but using [pytest-xdist](https://github.com/pytest-dev/pytest-xdist) might be one way to avoid it?

NoraLoose:
With `pytest-xdist` there are no errors anymore on Perlmutter, but the pytest suite simply stops after 33% of the tests are executed.

NoraLoose:
Update on this issue:
* After merging #121 and #128, which enabled running the tests without dask, this was no longer an issue (unless you did `pytest --use_dask`)
* After merging #140, which got rid of numba, we can now also run `pytest --use_dask` on perlmutter without issues

To summarize, this was an issue related to the combination of `dask` and `numba` on perlmutter. Using each of them in isolation worked fine. Now `numba` is not used within `ROMS-Tools` anymore, so this is no longer an issue, and I'm closing it.

PR #121, opened 2024-09-11
Make dask optional

NoraLoose:
This PR makes the usage of dask optional via a `use_dask` parameter (with default `True`) for each of the following classes.
* `BoundaryForcing`
* `InitialConditions`
* `SurfaceForcing`
* `TidalForcing`

Note that the grid generation process is currently not using dask ever, so we are not giving the `use_dask` parameter to the `Grid` class.

I also adapted the test suite to optionally use dask or not.
```
pytest
```

will run all tests without dask, while
```
pytest --use_dask
```
will run all tests with dask.

cc @TomNicholas @dafyddstephenson

### DIFF
--- a/roms_tools/setup/datasets.py
+++ b/roms_tools/setup/datasets.py
@@ -128,6 +128,7 @@ class Dataset:
         """
 
         # Check for wildcards in the filename
+        # xarray.open_mfdataset is currently only working with chunks, see https://github.com/pydata/xarray/issues/9038
         if not self.use_dask and re.search(r"[\*\?\[\]]", self.filename):
             raise ValueError(
                 "Wildcards detected in the filename but use_dask=False. "
--- a/roms_tools/setup/datasets.py
+++ b/roms_tools/setup/datasets.py
@@ -1,3 +1,4 @@
+import re
 import xarray as xr
 from dataclasses import dataclass, field
 import glob
@@ -23,7 +24,7 @@ class Dataset:
     Parameters
     ----------
     filename : str
-        The path to the data files. Can contain wildcards.
+        The path to the data files. Can contain wildcards if use_dask is True.
     start_time : Optional[datetime], optional
         The start time for selecting relevant data. If not provided, the data is not filtered by start time.
     end_time : Optional[datetime], optional
@@ -122,7 +123,16 @@ class Dataset:
         ------
         FileNotFoundError
             If the specified file does not exist.
+        ValueError
+            If wildcards are found in the filename and use_dask=False.
         """
+
+        # Check for wildcards in the filename
+        if not self.use_dask and re.search(r"[\*\?\[\]]", self.filename):
+            raise ValueError(
+                "Wildcards detected in the filename but use_dask=False. "
+                "Wildcards can only be used when use_dask=True."
+            )
 
         # Check if the file exists
         matching_files = glob.glob(self.filename)
--- a/roms_tools/setup/datasets.py
+++ b/roms_tools/setup/datasets.py
@@ -162,7 +162,6 @@ class Dataset:
                         coords="minimal",
                         compat="override",
                         chunks=chunks,
-                        engine="netcdf4",
                     )
                 else:
                     ds = xr.open_dataset(
--- a/roms_tools/setup/datasets.py
+++ b/roms_tools/setup/datasets.py
@@ -25,7 +25,7 @@ class Dataset:
     Parameters
     ----------
     filename : Union[str, Path, List[Union[str, Path]]]
-        The path to the data file(s). Can be a single string (with or without wildcards)
+        The path to the data file(s). Can be a single string (with or without wildcards), a single Path object,
         or a list of strings or Path objects containing multiple files.
     start_time : Optional[datetime], optional
         The start time for selecting relevant data. If not provided, the data is not filtered by start time.
--- a/roms_tools/setup/datasets.py
+++ b/roms_tools/setup/datasets.py
@@ -4,7 +4,8 @@ from dataclasses import dataclass, field
 import glob
 from datetime import datetime, timedelta
 import numpy as np
-from typing import Dict, Optional
+from typing import Dict, Optional, Union, List
+from pathlib import Path
 import dask
 import warnings
 from roms_tools.setup.utils import (
@@ -23,8 +24,9 @@ class Dataset:
 
     Parameters
     ----------
-    filename : str
-        The path to the data files. Can contain wildcards if use_dask is True.
+    filename : Union[str, Path, List[Union[str, Path]]]
+        The path to the data file(s). Can be a single string (with or without wildcards)
+        or a list of strings or Path objects containing multiple files.
     start_time : Optional[datetime], optional
         The start time for selecting relevant data. If not provided, the data is not filtered by start time.
     end_time : Optional[datetime], optional
@@ -60,7 +62,7 @@ class Dataset:
     Dimensions:  ...
     """
 
-    filename: str
+    filename: Union[str, Path, List[Union[str, Path]]]
     start_time: Optional[datetime] = None
     end_time: Optional[datetime] = None
     var_names: Dict[str, str]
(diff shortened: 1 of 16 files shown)


NoraLoose:
It remains to be seen if this solves #61 on Perlmutter. Perlmutter is currently down for maintenance.

roms_tools/setup/datasets.py
```diff
@@ -118,8 +123,17 @@ def load_data(self) -> xr.Dataset:
...
+        ValueError
+            If wildcards are found in the filename and use_dask=False.
         """
 
+        # Check for wildcards in the filename
+        if not self.use_dask and re.search(r"[\*\?\[\]]", self.filename):
+            raise ValueError(
+                "Wildcards detected in the filename but use_dask=False. "
+                "Wildcards can only be used when use_dask=True."
+            )
```
TomNicholas, line 135: Should probably leave a note for devs here that links to https://github.com/pydata/xarray/issues/9038

roms_tools/setup/datasets.py
```diff
@@ -128,31 +142,39 @@ def load_data(self) -> xr.Dataset:
...
+                    chunks[self.dim_names["time"]] = 1
+
+                    ds = xr.open_mfdataset(
+                        self.filename,
+                        combine="nested",
+                        concat_dim=self.dim_names["time"],
+                        coords="minimal",
+                        compat="override",
+                        chunks=chunks,
+                        engine="netcdf4",
```
TomNicholas, line 164: Is there a reason why we pass the `engine` kwarg in one case but not the other?

roms_tools/setup/datasets.py
```diff
@@ -23,7 +24,7 @@ class Dataset:
     Parameters
     ----------
     filename : str
-        The path to the data files. Can contain wildcards.
+        The path to the data files. Can contain wildcards if use_dask is True.
```
TomNicholas, line 27: There's a potential trap here - wildcards sort alphabetically, and because we're using `combine='nested'` then the datasets will be concatenated in the order given. So if our datasets are ever stored under paths which shouldn't be concatenated in alphabetical order then this will give an unintended result.

roms_tools/tests/test_setup/test_boundary_forcing.py
```diff
@@ -12,6 +12,7 @@ def test_boundary_forcing_creation(boundary_forcing):
     Test the creation of the BoundaryForcing object.
     """
 
+    print(boundary_forcing)
```
TomNicholas, line 15:
```suggestion
```

roms_tools/tests/test_setup/conftest.py
```diff
@@ -18,6 +18,9 @@ def pytest_addoption(parser):
         default=[],
         help="Specify which fixtures to overwrite. Use 'all' to overwrite all fixtures.",
     )
+    parser.addoption(
+        "--use_dask", action="store_true", default=False, help="Run tests with Dask"
+    )
```
TomNicholas, line 23: It would be good to run the tests once with and once without dask, in different CI jobs.

TomNicholas (approved):
Nice!

roms_tools/setup/datasets.py
NoraLoose: Removed the `engine` kwarg in `1df2b65`.

roms_tools/setup/datasets.py
NoraLoose: Done in [d6db2f6](https://github.com/CWorthy-ocean/roms-tools/pull/121/commits/d6db2f6665403778ba466b928dd3e3daa3e7d3f3)

roms_tools/setup/datasets.py
NoraLoose: Good point. Can you think of a better solution in terms of the kwargs that we give to `xr.open_mfdataset`?

TomNicholas: 1. Use `combine='by_coords'` (only works if there are coordinates to use to combine, but for `time` there should be),
2. Just make users pass a `list[str]` instead of a wildcard. Then there are no surprises, and that list should be passed directly on to `xr.open_mfdataset`. 
 
I think I would prefer option (2) as it feels less likely to cause any surprises.

roms_tools/tests/test_setup/conftest.py
NoraLoose: Good idea. As of [f3e9c5e](https://github.com/CWorthy-ocean/roms-tools/pull/121/commits/f3e9c5e80aef826bd45911485c6f113c98b8e174), the tests are run with and without dask in the CI. This doubles the CI test suite run time.

roms_tools/setup/datasets.py
NoraLoose: The problem with (2) is that the list may be very long. For example, the GLORYS typically has one file per day. If the user wants to produce input data for a year, they have to make a list that consists of 365 items. @dafyddstephenson do you have thoughts on this?

roms_tools/tests/test_setup/conftest.py
TomNicholas: Are these getting run in the same CI jobs? I don't see them in the list below. Having a separately listed job would be useful for quickly diagnosing the cause of a failure I think.

We could ask `--use-dask` to the matrix, and test every combination of python version with and without dask. But that seems overkill. Perhaps we should run most jobs without (with?) dask, and one one extra job with (without?) dask.

NoraLoose: They are being run within the same CI job.
```
      - name: Running Tests (without Dask)
        shell: bash -l {0}
        run: |
           python -V
           coverage run --rcfile=coverage.toml -m pytest --verbose roms_tools/tests/*
      - name: Running Tests (with Dask)
        shell: bash -l {0}
        run: |
           python -V
           coverage run --rcfile=coverage.toml -m pytest --use_dask --verbose roms_tools/tests/*
```
but I can make a separate job.

NoraLoose: Done! The CI tests now include:
* 3.10-build-without-dask
* 3.11-build-without-dask
* 3.12-build-without-dask
* 3.12-build-with-dask

roms_tools/setup/datasets.py
NoraLoose: Would you be okay with (1) @TomNicholas? I just head from @dafyddstephenson that he would prefer (1) as well. 

Just to clarify, if we go for (1), we would change
```
                    ds = xr.open_mfdataset(
                        self.filename,
                        combine="nested",
                        concat_dim=self.dim_names["time"],
                        coords="minimal",
                        compat="override",
                        chunks=chunks,
                    )
```
to
```
                    ds = xr.open_mfdataset(
                        self.filename,
                        combine="by_coords",
                        coords="minimal",
                        compat="override",
                        chunks=chunks,
                    )
```
???

roms_tools/tests/test_setup/conftest.py
TomNicholas: Amazing, thanks!

roms_tools/setup/datasets.py
TomNicholas: Yes - but maybe just double-check that the additional checking done by `"by_coords"` doesn't significantly slow down the opening.

NoraLoose: Good point, now I remember. This performance issue is the reason I went for 
```
                        combine="nested",
                        concat_dim=self.dim_names["time"],
```
in the first place.

If you have 365 GLORYS files sitting in your directory, `combine="by_coords"` slows the process down by more than 40 seconds, while the prior solution is almost instant.

NoraLoose: We could support both (1) and (2). 

(1) If the user wants to use wildcards, we open with `combine="by_coords"` but the user has to be fine with a significant performance hit.
(2) If the user provides a list, we open with `combine="nested", concat_dim=self.dim_names["time"]`, and it will be faster.

TomNicholas: It's a little more confusing than just having one option but I think thats fine.

NoraLoose:
Can we merge this @TomNicholas?

TomNicholas:
The only other thing I can think to add might be to make the non-dask tests actually use a different environment - one that doesn't even have dask installed. Then there is no way that you can accidentally use dask without realising.

NoraLoose:
Sure, we can try that. So how do you force conda to not install `dask` given this yaml?

https://github.com/CWorthy-ocean/roms-tools/blob/ae42e3be9d7d6b41d7bf3268e28f05d522d5996c/ci/environment.yml#L1-L30

TomNicholas:
I don't know 😶 At least one of those conda packages must have dask as a required dependency...

NoraLoose:
Dang, I think it might be gcm-filters. I opened issue #125 so we remember to do this in the future. Let's merge this for now?

TomNicholas:
Sounds good!

PR #128, opened 2024-09-13
Add PyYAML as required dependency

NoraLoose:
Now that the newest version of GCM-Filters (v0.5.0) does not require dask anymore, `PyYAML` (which is a dependency of dask) does not get installed anymore. We need to add it explicitly as a dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,8 @@ dependencies = [
     "packaging",
     "scipy",
     "gcm_filters",
-    "numba"
+    "numba",
+    "PyYAML"
 ]
 
 dynamic = ["version"]


NoraLoose:
Actually, the only time we explicitly use dask is this single line:

https://github.com/CWorthy-ocean/roms-tools/blob/8f0faac33dc62e72d37cd9e9bc88a8ea4d951004/roms_tools/setup/datasets.py#L176

If we get rid off it (as @TomNicholas suggested before), we can avoid importing dask altogether.

NoraLoose:
```
Collecting roms-tools[dask]
  Downloading roms_tools-1.2.0-py3-none-any.whl.metadata (2.9 kB)
WARNING: roms-tools 1.2.0 does not provide the extra 'dask'
```
is concerning

NoraLoose:
@TomNicholas I would like to merge this asap because the removed dask import will solve some problems with C-Star that @dafyddstephenson was facing.

NoraLoose:
I updated the description for what this PR is doing (see first comment) - it has evolved quite a bit.

TomNicholas (approved):
Excellent

Issue #70, opened 2024-08-13
Avoid extrapolation to bottom before interpolation


NoraLoose:
Jeroen strongly advises against extrapolating tracers and velocities all the way to the bottom before interpolation. Instead, he recommends to do horizontal interpolation (or diffusion). His recommendation is the solution that I had implemented first, but then we switched to extrapolating to the bottom.

The argument against extrapolation to the bottom is that you can end up with very warm temperatures at the ocean bottom if the bathymetry is slightly shifted. The warm temperature will be next to cold (realistic) temperatures, and there will therefore be strong gradients that may require initially very small time steps (that the user may miss to increase later on). In contrast, horizontal interpolation will supposedly lead to no problems because at depth isopycnals are almost horizontal (as opposed to the upper ocean where isopycnals are much more tilted). The case of flat isopycnals is the limit case for when horizontal interpolation is perfect.

matt-long:
@nmolem, what is the rationale here? I would have expected convectively unstable conditions to be the primary concern.

nmolem:
The rationale is extrapolation in horizontal instead of vertical direction. Since the vertical gradients in the ocean are typically much larger, it makes more sense to extrapolate horizontally. Bathymetry differences, even between ROMS grids with different resolutions can be 1000m or more. If you extrapolate the lighter waters downward by that distance, you create large horizontal density gradients, which need a small time-step to get pas the transients. With the method that I use, typically, there is no need for a smaller time-step after initialization. In the (rare) case that you end up with statically unstable gradients, the Ri-number parameterization will erode those efficiently.

nmolem:
I initially was extrapolating vertically, because it's the simplest thing to do, but switched to the more complicated horizontal extrapolation to avoid the worst transients.

PR #140, opened 2024-09-23
Algebraic Multigrid (AMG) method for filling land values

NoraLoose:
This PR substitutes the SOR method for a much more efficient Algebraic Multigrid (AMG) method to propagate ocean values into land, enabled by a new dependency, [`PyAMG`](https://github.com/pyamg/pyamg). I took @sdbachman' code (see #134) which operates on 2D numpy arrays and vectorized it via `xarray.apply_ufunc` so that we 
* don't have to loop over time slices and vertical layers
* can take advantage of dask for parallelization

## Reproducibility Issue
It seems that the AMG method is not entirely deterministic. (Can you confirm @bensworth?) When you set up a solver twice, there will be tiny differences. These tiny differences are detected by our tests that check for bit-by-bit reproducible results via file hashes (these are the tests that fail below). The tests that check for **close** results (up to a small epsilon) pass. We can either remove the tests that check for bit-by-bit reproducibility, or contribute a solution to PyAMG that enables bit-by-bit reproducible results, e.g. by exposing a random seed function as suggested in https://github.com/pyamg/pyamg/issues/350.

## Future Work
This PR assumes a single 2D mask (for 3D data like GLORYS inferred from the surface), and generates a single solver per dataset. This is okay for now because for 3D data we extrapolate values from the surface to the bottom before we use AMG to fill land values horizontally. However, we are probably planning to move away from the vertical extrapolation in the future (see discussion in #70). In that case, we need a 3D mask, and a different solver for each depth level.


NoraLoose:
In order to make the tests pass, I had to take two steps:
* Comment out the tests that check for bit-by-bit reproducible results via file hashes.
* Increase the relative tolerance rtol in the validation tests from 1e-5 to 1e-2.

We want to revert both of these steps once the reproducibility problem is fixed upstream in PyAMG. Let's leave this for future work (I opened issue #142). I will merge this PR.

PR #145, opened 2024-09-24
Enable opening of multiple files without dask

### LINKED ISSUE #143: Open multiple files without dask
When the user creates an object
```
Object(
    grid=grid,
    source={"name": name, "path": path},
    use_dask=use_dask
)
```
`path` can be a 
- [ ] list of filenames
- [ ] wildcards

if `use_dask = True`. However, if `use_dask = False`, these two options are currently not available - instead the user has to provide a single filename. 
https://github.com/CWorthy-ocean/roms-tools/blob/0a1d0cd268d1b5f35e9f6529f4b40ff9c6699b14/roms_tools/setup/datasets.py#L177-L201
This is because `xr.open_mfdataset` is not supported without dask.

There is a work-around, though. We can loop over multiple files, open them via `xr.open_dataset` and the concatenate them, resulting in eagerly loaded data.

NoraLoose:
Closes #143.

### DIFF
--- a/roms_tools/setup/datasets.py
+++ b/roms_tools/setup/datasets.py
@@ -183,7 +183,7 @@ class Dataset:
             kwargs = {"combine": "nested", "concat_dim": self.dim_names["time"]}
 
         # Base kwargs used for dataset combination
-        kwargs0 = {
+        combine_kwargs = {
             "coords": "minimal",
             "compat": "override",
             "combine_attrs": "override",
@@ -203,7 +203,7 @@ class Dataset:
             ds = xr.open_mfdataset(
                 matching_files,
                 chunks=chunks,
-                **kwargs0,
+                **combine_kwargs,
                 **kwargs,
             )
         else:
@@ -213,10 +213,10 @@ class Dataset:
                 ds_list.append(ds)
 
             if kwargs["combine"] == "by_coords":
-                ds = xr.combine_by_coords(ds_list, **kwargs0)
+                ds = xr.combine_by_coords(ds_list, **combine_kwargs)
             elif kwargs["combine"] == "nested":
                 ds = xr.combine_nested(
-                    ds_list, concat_dim=kwargs["concat_dim"], **kwargs0
+                    ds_list, concat_dim=kwargs["concat_dim"], **combine_kwargs
                 )
 
         if "time" in self.dim_names and self.dim_names["time"] not in ds.dims:
--- a/roms_tools/setup/datasets.py
+++ b/roms_tools/setup/datasets.py
@@ -184,7 +184,11 @@ class Dataset:
             kwargs = {"combine": "nested", "concat_dim": self.dim_names["time"]}
 
         # Base kwargs used for dataset combination
-        kwargs0 = {"coords": "minimal", "compat": "override"}
+        kwargs0 = {
+            "coords": "minimal",
+            "compat": "override",
+            "combine_attrs": "override",
+        }
 
         if self.use_dask:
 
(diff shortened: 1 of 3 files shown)


roms_tools/setup/datasets.py
```diff
@@ -174,6 +174,21 @@ def load_data(self) -> xr.Dataset:
...
+        # Determine the kwargs for combining datasets
+        if re.search(r"[\*\?\[\]]", str(filename_str)) or len(matching_files) == 1:
+            # If there is a wildcard or just one file, use by_coords
+            kwargs = {"combine": "by_coords"}
+        else:
+            # Otherwise, use nested combine based on time
+            kwargs = {"combine": "nested", "concat_dim": self.dim_names["time"]}
+
+        # Base kwargs used for dataset combination
+        kwargs0 = {
```
TomNicholas, line 186: Rename to `combine_kwargs` or something

roms_tools/setup/datasets.py
```diff
@@ -160,12 +154,18 @@ def load_data(self) -> xr.Dataset:
...
         # Handle the case when filename is a list
         elif isinstance(filename_str, list):
-            if not self.use_dask:
-                raise ValueError(
-                    "A list of files is provided, but use_dask=False. "
-                    "Multiple files can only be loaded when use_dask=True."
-                )
-            matching_files = filename_str
+            # Check if any element in the list contains wildcard characters
+            if any(re.search(r"[\*\?\[\]]", f) for f in filename_str):
```
TomNicholas, line 158: You do the same regex search more than once so you might be able to cache it using a different function from the regex std lib.

roms_tools/setup/datasets.py
```diff
@@ -174,6 +174,21 @@ def load_data(self) -> xr.Dataset:
                 "A time dimension must be available to concatenate the files."
             )
 
+        # Determine the kwargs for combining datasets
+        if re.search(r"[\*\?\[\]]", str(filename_str)) or len(matching_files) == 1:
+            # If there is a wildcard or just one file, use by_coords
+            kwargs = {"combine": "by_coords"}
+        else:
+            # Otherwise, use nested combine based on time
+            kwargs = {"combine": "nested", "concat_dim": self.dim_names["time"]}
```
TomNicholas, line 183: A double-nested list is another valid argument to `combine_nested` that would probably go wrong if passed here.

TomNicholas (approved):
Looks good with a brief look just now. I can look more tomorrow if helpful.

roms_tools/setup/datasets.py
NoraLoose: But we will never create a double-nested list with the current implementation, correct?

roms_tools/setup/datasets.py
NoraLoose: Good point. Done in 22e7b7c.

roms_tools/setup/datasets.py
NoraLoose: Done in aba43d3.

roms_tools/setup/datasets.py
TomNicholas: If a double-nested list can never get here then no problem, you're right. (I only skimmed this on the subway!)

NoraLoose: Haha, nice! It can only get here if the user provides a double-nested list, but we require `filename : Union[str, Path, List[Union[str, Path]]]`, which excludes that case.

NoraLoose:
I think we can merge this.

PR #165, opened 2024-10-16
Fix mask at boundaries

### LINKED ISSUE #163: Grid mask incorrect for points near edge
image

See the above image - near the northeast corner of the domain, ROMS has added an additional masked point at the edge based on the masked point in the interior. This behaviour should be the same in `roms-tools`.

NoraLoose:
Closes #163.


NoraLoose:
The failing tests are not introduced by this PR. I opened a separate issue #166.

Issue #193, opened 2024-11-13
Segmentation fault when running tests with Dask on Github Actions


NoraLoose:
PR #192 simply changes the documentation while leaving the code base untouched. Still, the tests fail, but only when using dask:

https://github.com/CWorthy-ocean/roms-tools/actions/runs/11824684038/job/32946744730?pr=192

It looks like dask may be consuming too many resources on GitHub Actions, leading to a segmentation fault. Not sure what is going on. The tests with dask pass just fine locally.

@TomNicholas?

NoraLoose:
@ankona do you mind posting the NetCDF failure in the tests with Dask here for the record?

PR #195, opened 2024-11-20
River forcing addendum

NoraLoose:
This is a small follow-up PR to #194. It 

* makes sure that the case if no relevant rivers are found is handled correctly
* adds more unit tests


NoraLoose:
The failing test is due to a separate issue (#193). I'm merging this.

Issue #201, opened 2024-12-04
Test and document `ROMS-Tools` for large / operational use cases


NoraLoose:
Until now, we have used `ROMS-Tools` for small test cases. Next, we need to test it on large and operational use cases, which might involve the correct usage of Dask clusters etc.

Here is a checklist for what object have been tested for operational use cases:

- [x] Grid / Topography / Mask (use case: 5000 x 5000 grid documented [here](https://roms-tools.readthedocs.io/en/latest/grid.html#Topography))
- [ ] Tidal Forcing
- [ ] Surface Forcing
- [ ] Initial Conditions
- [ ] Boundary Forcing
- [ ] River Forcing

NoraLoose:
# Testing on Perlmutter

## Jupyter Notebook run on CPU node (~250GB of memory)

### Grid / Topography / Mask (`Grid` object)

####  Grid A: 

- [x] `Grid` initialization:  peak memory: 701.46 MiB, increment: 502.90 MiB, Wall time: 12.7 s
- [x] `grid.plot_vertical_coordinate(s=s)`: peak memory: 2959.17 MiB, increment: 2384.32 MiB, Wall time: 1.47 s
- [x]  `grid.plot_vertical_coordinate(eta=eta)`: peak memory: 584.21 MiB, increment: 0.60 MiB, Wall time: 294 ms


### Tidal Forcing
### Surface Forcing
### Initial Conditions
### Boundary Forcing
### River Forcing

NoraLoose:
# Testing on Perlmutter

* Jupyter Notebook run on CPU node (~500GB? of memory)
* use Dask in default thread-based mode

## Tidal Forcing

### Grid A

Memory footprint of `tidal_forcing.ds`: `374.46 MB`

#### Without Dask

- [x] `TidalForcing` initialization, `bypass_validation = False`:
```
peak memory: 1887.27 MiB, increment: 1322.11 MiB
CPU times: user 1min 10s, sys: 1.34 s, total: 1min 11s
Wall time: 4.05 s
```

- [x] `TidalForcing` initialization, `bypass_validation = True`:
```
peak memory: 1934.29 MiB, increment: 1397.38 MiB
CPU times: user 1min 10s, sys: 1.15 s, total: 1min 11s
Wall time: 4.02 s
```

#### With Dask

- [x] `TidalForcing` initialization, `bypass_validation = False`:
```
peak memory: 1213.72 MiB, increment: 671.60 MiB
[...]

PR #221, opened 2025-01-23
Remove support for spatial partitioning upon saving

NoraLoose:
## Changes

This PR removes the support for spatial partitioning upon saving—a functionality that, to my knowledge, is not actively used, not by C-Star at least. (Note that the user can still use ROMS-Tools to partition their *existing* files, via the `partition_netcdf` function.) The rationale of this PR is a performance gain, both in terms of faster compute times and memory savings. 

- [x] Tests removed that test partitioning upon saving
- [x] Functions/methods are updated in `docs/api.rst`

## Issue resolved

To enable spatial partitioning we had to load the dataset into memory first before they were spatially partitioned and then saved to disk, see https://github.com/CWorthy-ocean/roms-tools/pull/114. This means we could not take advantage of Dask's optimized I/O capabilities. As of this PR, we never load the datasets into memory (when `use_dask = True`), they will simply be written to disk by Dask, which results in an improved memory footprint and faster computations.

## Example of performance gain

Saving 1 month of hourly surface forcing data for grid A in https://github.com/CWorthy-ocean/roms-tools/issues/201 with `use_dask = True` before and after this PR:

Before:
```
peak memory: 82923.36 MiB, increment: 82109.48 MiB
CPU times: user 7h 14min 36s, sys: 2min, total: 7h 16min 37s
Wall time: 8min 35s
```
After:
```
peak memory: 67603.79 MiB, increment: 66888.29 MiB
CPU times: user 6h 56min 22s, sys: 2min 20s, total: 6h 58min 42s
Wall time: 3min 37s
```
For context: Memory footprint of actual dataset: 23634.59 MB


NoraLoose:
The tests with Dask fail, due to issue #222. I am still merging this PR, and I'm hoping we can tackle #222 separately.

Issue #282, opened 2025-04-01
Upload to PyPI failing


NoraLoose:
I just released a new version (v2.6.0), but the automated upload to PyPI is failing:

https://github.com/CWorthy-ocean/roms-tools/actions/runs/14203050792/job/39795374971

The error message is
```
ERROR    InvalidDistribution: Metadata is missing required fields: Name,        
         Version.                                                               
         Make sure the distribution includes the files where those fields are   
         specified, and is using a supported Metadata-Version: 1.0, 1.1, 1.2,   
         2.0, 2.1, 2.2.                                                         
```

The error message indicates that the version cannot be determined correctly, which is unexpected since this process has worked previously. Indeed, 
```
python -m setuptools_scm
```
gives me
```
2.6.0
```
as expected.

The big change for this release is that I added `xgcm` and `numba` as dependencies in the `pyproject.toml`. How can this possibly have broken things?

@ScottEilerman @dafyddstephenson @ankona

NoraLoose:
Closed via #285

PR #283, opened 2025-04-01
Rename `river_location` to `river_flux`

NoraLoose:
This is a small PR that renames `river_location` to `river_flux` for the sake of (current) compatibility with ROMS.


NoraLoose:
I want to solve #282 first before merging this PR.

PR #285, opened 2025-04-02
Update pypi package action, add test upload action

ankona:
This PR contains the following changes:

1. Fix an issue with publishing to PyPi using an old version of the Github `pypa/gh-action-pypi-publish` action.
2. Add a Github action for uploading tagged commits to test pypi

### DIFF
--- a/.github/workflows/pypipublish.yaml
+++ b/.github/workflows/pypipublish.yaml
@@ -65,6 +65,27 @@ jobs:
           python -m pip install dist/roms_tools*.whl
           python -c "import roms_tools; print(roms_tools.__version__)"
 
+  upload-to-test-pypi:
+    needs: test-built-dist
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    runs-on: ubuntu-latest
+    environment:
+      name: testpypi
+      url: https://test.pypi.org/p/roms-tools
+
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: releases
+          path: dist
+      - name: Publish package to TestPyPI
+        uses: pypa/gh-action-pypi-publish@v1.12.4
+        with:
+          repository-url: https://test.pypi.org/legacy/
+
   upload-to-pypi:
     needs: test-built-dist
     if: github.event_name == 'release'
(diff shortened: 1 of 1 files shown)


.github/workflows/pypipublish.yaml
```diff
@@ -78,6 +99,6 @@ jobs:
           name: releases
           path: dist
       - name: Publish package to PyPI
-        uses: pypa/gh-action-pypi-publish@v1.8.10
+        uses: pypa/gh-action-pypi-publish@v1.12.4
```
NoraLoose, line 102: Wow, that might do the trick!

.github/workflows/pypipublish.yaml
```diff
@@ -65,6 +65,27 @@ jobs:
           python -m pip install dist/roms_tools*.whl
           python -c "import roms_tools; print(roms_tools.__version__)"
 
+  upload-to-test-pypi:
+    needs: test-built-dist
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
NoraLoose, line 70: Does it make sense to loosen this constraint?

NoraLoose: It would be great to uncover a potential uploading to PyPI issue before merging the PR. Is that possible?

NoraLoose:
Apart from the question that I posted above, this looks great to me. Feel free to merge @ankona.

.github/workflows/pypipublish.yaml
ankona: I removed the constraint, altogether. I misunderstood the amount of manual commits to main, so the existing constraints on the whole step should do the trick. I am curious how `setuptools_scm` generates version numbers without a release. Might need a tweak, still.

ankona: Moving test pypi release to a new ticket to unblock this

NoraLoose:
This is not the first time I have seen 
```
ERROR roms_tools/tests/test_setup/test_boundary_forcing.py::test_files_have_same_hash - RuntimeError: NetCDF: HDF error
```

It usually works to re-trigger the check, but that's not a good long-time solution. We should add it to [this issue](https://github.com/CWorthy-ocean/roms-tools/issues/193), and investigate it sooner or later.

ankona:
> This is not the first time I have seen
> 
> ```
> ERROR roms_tools/tests/test_setup/test_boundary_forcing.py::test_files_have_same_hash - RuntimeError: NetCDF: HDF error
> ```
> 
> It usually works to re-trigger the check, but that's not a good long-time solution. We should add it to [this issue](https://github.com/CWorthy-ocean/roms-tools/issues/193), and investigate it sooner or later.
> 
> Edit: It always only shows up in combination with dask. 😕

I'm re-running just to be safe. If it fails, I'll merge so we can move on.

PR #298, opened 2025-04-29
Introduce Pydantic `Release` object

### LINKED ISSUE #296: Refactor `CDRVolumePointSource.add_release` to Use Pydantic `Release` Object
For a future refactor, we could introduce a `Release` object using Pydantic to handle input validation for the `CDRVolumePointSource.add_release` method. This would simplify the code, eliminate the need for custom validation logic, and streamline data handling.

Original comment: https://github.com/CWorthy-ocean/roms-tools/pull/295#discussion_r2059079189.

NoraLoose:
This PR refactors the `CDRVolumePointSource` and introduces a Pydantic `Release` object, enabling:

- Cleaner code by avoiding repetition of parameters like `lat`, `lon`, and `depth` in method calls  
- Built-in and custom validation through Pydantic

It also updates regex patterns in the tests to align with the new Pydantic validation error messages.

- [x] Closes #296
- [x] Tests added


roms_tools/setup/cdr_forcing.py
```diff
@@ -28,6 +29,195 @@
...
+    - Validates that `lat`, `depth`, `times`, `volume_fluxes`, and `tracer_concentrations` follow expected formats.
+    - Converts single values for `volume_fluxes` and `tracer_concentrations` into time-series matching `times`.
+    """
+
+    name: str
+    lat: Union[float, int] = Field(ge=-90, le=90)
+    lon: Union[float, int]
+    depth: Union[float, int] = Field(ge=0)
+    times: List[datetime]
+    volume_fluxes: Union[float, int, List[Union[float, int]]]
```
NoraLoose, line 67: I don't think that a built-in Pydantic constraint for non-negativity is feasible here due to the flexibility of the input type (either a single float/int or a list of floats/ints). Instead, non-negativity is enforced manually in the `@model_validator` method `check_and_convert_volume_fluxes` below.

I'm using the same approach is used for `tracer_concentrations`.

NoraLoose:
@ScottEilerman This is my first time working with Pydantic. Is this refactor more or less what you had in mind? Happy about feedback if you have a minute.

roms_tools/setup/cdr_forcing.py
ScottEilerman: You might be able to use these: https://docs.pydantic.dev/latest/concepts/types/#constrained-types

e.g. `volume_fluxes: confloat(ge=0) | conlist(confloat(ge=0), min_length=1)`

ScottEilerman: (As a general comment, recent versions of python support using type constructors and logic operators as type hints instead of needing stuff from the typing module, e.g. `List[foo]` -> `list[foo]`, Union[foo, bar] -> `foo | bar`)

NoraLoose: Great, thanks. I ended up defining
```python
from typing_extensions import Annotated
from annotated_types import Ge
NonNegativeFloat = Annotated[float, Ge(0)]
```
and then
```
volume_fluxes: Union[NonNegativeFloat, List[NonNegativeFloat]]
```
which seems to work, see [c3f1a67](https://github.com/CWorthy-ocean/roms-tools/pull/298/commits/c3f1a676bbeeca00baa887d3a8ba4aceae9a3b92).

NoraLoose: but maybe we should stay within pydantic, and not use `annotated_types`

NoraLoose: ChatGPT tells me:

✅ In Pydantic v2, constraints like Ge, Le, Gt, Lt, etc. are provided by the [annotated-types](https://pypi.org/project/annotated-types/) package (a separate but lightweight dependency).

✅ pydantic v2 uses standard Annotated types from typing or typing_extensions + annotated_types for validation.

🔄 In Pydantic v1, you would use Field(ge=0) directly inside the Field(...) function, but this isn't compatible with item-wise validation inside a List.

NoraLoose:
The failing check has nothing to do with this PR, but is a random error that we have seen before and would go away upon retriggering the check:
```
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://raw.githubusercontent.com/CWorthy-ocean/roms-tools-test-data/main/CESM_regional_test_data_one_time_slice.nc
```
I'm merging this PR, so that we can move on.

PR #301, opened 2025-05-07
Refactor CDR forcing with support for `VolumeRelease` and `TracerPerturbation`

### LINKED ISSUE #302: Support two types of CDR releases
We want to support two types of CDR releases:

- **Volume point sources**: Represent point sources that inject both water and biogeochemical (BGC) tracers at specified locations—such as through an underwater pipe.

  - Designed for field-scale deployments where localized mixing is critical.  
  - Introduces both volume flux and tracer concentrations at fixed points.  
  - Requires all 32 MARBL BGC tracers, as well as temperature and salinity (defaults are provided if not specified).

- **Tracer perturbations**: Apply tracer-only perturbations to existing BGC fields without introducing any water (i.e., no volume flux).

  - Intended for large-scale simulations where small-scale mixing processes are implicitly represented.  
  - Only the specified tracers are altered; all others remain unchanged.

NoraLoose:
### Summary

This PR refactors the core logic for constructing Carbon Dioxide Removal (CDR) forcings compatible with ROMS. It addresses #302 along with its sub-issues:  #303, #304

#### **Release**

- Moves the base `Release` class into its own dedicated module.
- Introduces two concrete subclasses:
  - `VolumeRelease`: Represents a CDR release defined by a volume flux and associated tracer concentrations.
  - `TracerPerturbation`: Represents a tracer-only perturbation with tracer fluxes and without any volume flux.

These subclasses (`VolumeRelease` and `TracerPerturbation`) are now part of the public API. Users are expected to construct them explicitly before building a `CDRForcing` object (see documentation notebook for examples).

#### **New Core Classes** ...
... in the spirit of Composition over Inheritance - thanks @ankona 
- `ReleaseSimulationManager`: Validates and adjusts a single release against the simulation time window and grid.
- `ReleaseCollector`: Collects and validates multiple releases against each other
- `CDRForcingDatasetBuilder`: Constructs the xarray `Dataset` to be saved as NetCDF.
- `CDRForcing`: A new top-level interface that replaces `CDRVolumePointSource.add_release(...)`.  All releases are now passed during instantiation. Adds a `.plot_tracer_flux()` method for visualization, relevant for the new `TracerPerturbation`.

### Other Enhancements

- **Refactored time handling**:
  - `convert_to_roms_time` split into:
    - `convert_to_relative_days` (general utility).
    - `add_time_info_to_ds` (specific to Dataset construction).
- **Improved YAML serialization**:
  - Better separation of `_to_dict` and `_to_yaml` now allow forcing-specific customizations (e.g., for `RiverForcing`, `CDRForcing`).
- **Tracer metadata**:
  - Metadata now supports optional `flux_units` via the `with_flux_units` flag.

---

- [x] Closes #302
- [x] Tests added


NoraLoose:
Since `VolumeRelease` and `TracerPerturbation` are subclasses of the Pydantic `Release`, all these Pydantic-specific methods show up in our API ref, which I find distracting to the user:
Screenshot 2025-05-15 at 7 32 04 PM

roms_tools/setup/river_forcing.py
```diff
@@ -674,7 +675,24 @@ def to_yaml(self, filepath: Union[str, Path]) -> None:
             The path to the YAML file where the parameters will be saved.
         """
 
-        _to_yaml(self, filepath)
+        # Serialize object into dictionary
+        yaml_data = _to_dict(self)
```
ankona, line 679: `yaml_data` is not yaml. consider renaming to something like `forcing_dict` and there is no need to tell me the output of `_to_dict` is a dictionary with an extra comment.

roms_tools/setup/river_forcing.py
```diff
@@ -674,7 +675,24 @@ def to_yaml(self, filepath: Union[str, Path]) -> None:
...
+        # Convert indices format
+
+        indices_data = yaml_data["RiverForcing"]["indices"]
+        serialized_indices = {
+            "_convention": "eta_rho, xi_rho"
+        }  # Add convention metadata
+
+        for key, value in indices_data.items():
```
ankona, line 687: Feel free to get rid of the comment stating "comma-separated string." If you don't think it's obvious, again consider renaming your variables to avoid needing a comma? e.g.

```
csv_indices = [f"{v0},{v1}" for (v0,v1) in value]
yaml_data["RiverForcing"]["indices"] = csv_indices
```

alternatively, consider NOT formatting here and instead performing formatting in a single place (the yaml serializer). so, here, we would ideally end up only creating the data _to be serialized_. The strength of this approach is that this method doesn't produce values that are tightly coupled to the consumer code.

```
yaml_data["RiverForcing"]["indices"] = value
```

roms_tools/setup/river_forcing.py
```diff
@@ -674,7 +675,24 @@ def to_yaml(self, filepath: Union[str, Path]) -> None:
...
+            serialized_indices[key] = [
+                f"{tup[0]}, {tup[1]}" for tup in value
+            ]  # Comma-separated string
+
+        yaml_data["RiverForcing"]["indices"] = serialized_indices
+
+        # Write to YAML
+        _to_yaml(yaml_data, filepath)
```
ankona, line 695: Consider renaming `_to_yaml` to `write_yaml` to avoid the extraneous comment. As-is, I expect to see a return value based on the naming pattern previously established (e.g. `to_dict` returns a dict)

roms_tools/setup/cdr_forcing.py
```diff
@@ -21,171 +18,210 @@
...
+        return self
 
-            if isinstance(val, list):
-                if len(val) != num_times:
-                    raise ValueError(
-                        f"The length of tracer '{tracer}' ({len(val)}) does not match the number of times ({num_times})."
-                    )
+    def determine_release_type(self) -> "ReleaseCollector":
+        """Ensure all releases are of the same type, and set the release_type."""
+        if not all(
```
ankona, line 97: consider a set trick to make it clearer that we're interested in membership size:

this helps to avoid asking "wait, why am i looking at the 0-th item?" when you read this later on, and it lets you log the underlying issue (which someone will inevitably want to fix the bug). 

It also gives you something you can return without asking for type info again.


```
release_types = set(map(type, self.releases))
if len(release_types) > 1:
    type_list = ", ".join(release_types)
    raise ValueError(f"Not all releases have the same type. Received: {type_list}")

return release_types.pop()
```

roms_tools/setup/cdr_forcing.py
```diff
@@ -21,171 +18,210 @@
...
-                else [vals[0], vals[0]]
-                for tracer, vals in self.tracer_concentrations.items()
+    def __init__(self, releases, model_reference_date, release_type):
+        self.releases = releases
+        self.model_reference_date = model_reference_date
+        self.release_type = release_type
+
+    def build(self) -> xr.Dataset:
+        all_times = [np.array(r.times, dtype="datetime64[ns]") for r in self.releases]
+        union_times = np.unique(np.concatenate(all_times))
```
ankona, line 115: Could we avoid concatenation of numpy arrays by initializing with all times?

```
all_times = itertools.chain.from_iterable(r.times for r in self.releases)
unique_times = np.array(all_times).unique() # note: unique_times, not union_times
```

roms_tools/setup/cdr_forcing.py
```diff
@@ -21,171 +18,210 @@
...
+        union_rel_times = [
+            convert_to_relative_days(t, self.model_reference_date) for t in union_times
+        ]
```
ankona, line 118: Can we vectorize the relative time conversion?

```
ref = np.datetime64(model_reference_date, "ns")
day_delta = np.timedelta64(1, "D")

relative = union_times - ref
relative_days = relative / day_delta
```

This should be much faster since it acts on the entire array at once instead of converting each value individually

roms_tools/setup/cdr_forcing.py
```diff
@@ -21,171 +18,210 @@
...
+            "cdr_vsc": {
+                "long_name": "Vertical scale of CDR release",
+                "units": "meters",
+            },
+        }
+        for var, attrs in attr_map.items():
+            ds[var].attrs.update(attrs)
+
+        if self.release_type == VolumeRelease:
+            ds = add_tracer_metadata_to_ds(
```
ankona, line 162: Consider pulling each of your big cases (`if self.release_type == xxx`) into it's own method to simplify testing of each subsection and shorten this monster!

roms_tools/setup/cdr_forcing.py
```diff
@@ -21,171 +18,210 @@
...
+        ds["cdr_lat"] = ("ncdr", [r.lat for r in self.releases])
+        ds["cdr_dep"] = ("ncdr", [r.depth for r in self.releases])
+        ds["cdr_hsc"] = ("ncdr", [r.hsc for r in self.releases])
+        ds["cdr_vsc"] = ("ncdr", [r.vsc for r in self.releases])
+        ds = ds.assign_coords(
+            {"release_name": (["ncdr"], [r.name for r in self.releases])}
+        )
+
+        # Assign attributes
+        attr_map = {
```
ankona, line 133: consider pulling this out into a class method to shorten the monster!

```
@classmethod
def get_attr_map(cls, reference_date: int) -> t.Dict[str, t.Dict[str, str]]:
    return { ... attr_map here ... }
```

roms_tools/setup/cdr_forcing.py
```diff
@@ -21,171 +18,210 @@
...
-                        self.tracer_concentrations[k][-1]
+            elif self.release_type == TracerPerturbation:
+                for ntracer in range(ds.ntracers.size):
+                    tracer_name = ds.tracer_name[ntracer].item()
+                    ds["cdr_trcflx"].loc[
+                        {"ntracers": ntracer, "ncdr": ncdr}
+                    ] = np.interp(
+                        union_rel_times,
+                        rel_times,
+                        release.tracer_fluxes[tracer_name].values,
```
ankona, line 216: notice how the inside of the two loops are identical? break apart the parameterization of the problem and solving it to avoid this repetition.

```py
# set up the parameters of the problem...
if self.release_type == VolumeRelease:
 tracer_key = "cdr_tracer"
 tracer_data = release.tracer_concentrations
elif self.release_type == TracerPerturbation:
  tracer_key = "cdr_trcflx"
  tracer_data = release.tracer_fluxes

# solve using params
for ntracer in range(ds.ntracers.size):
    tracer_name = ds.tracer_name[ntracer].item()
    ds[tracer_key].loc[
        {"ntracers": ntracer, "ncdr": ncdr}
    ] = np.interp(
        union_rel_times,
        rel_times,
        tracer_data[tracer_name].values,
    )
```

now we can clearly see that we're really only doing _one_ action.

roms_tools/setup/cdr_forcing.py
```diff
@@ -210,291 +246,37 @@ class CDRVolumePointSource:
     start_time: datetime
     end_time: datetime
     model_reference_date: datetime = datetime(2000, 1, 1)
-    releases: Optional[dict] = field(default_factory=dict)
+    releases: List["Release"] = field(default_factory=list)
```
ankona, line 249: i am super inconsistent here... do we want to use `typing.List[foo]` or new style `list[foo]`?

roms_tools/setup/cdr_forcing.py
```diff
@@ -210,291 +246,37 @@ class CDRVolumePointSource:
...
-
-        # Convert times to datetime64[ns]
-        times = np.array(release.times, dtype="datetime64[ns]")
-
-        # Ensure reference date is also datetime64[ns]
-        ref = np.datetime64(self.model_reference_date, "ns")
-
-        # Compute model-relative times in days
-        rel_times = (times - ref) / np.timedelta64(1, "D")
+        for release in self.releases:
```
ankona, line 263: Very nice encapsulation!

roms_tools/setup/cdr_forcing.py
```diff
@@ -503,136 +285,227 @@ def plot_volume_flux(self, start=None, end=None, releases="all"):
...
         end = end or self.end_time
 
-        # Handle "all" releases case
-        if releases == "all":
-            releases = [k for k in self.releases if k != "_tracer_metadata"]
-        # Validate input for release names
-        self._validate_release_input(releases)
+        valid_release_names = [r.name for r in self.releases]
+
+        if release_names == "all":
```
ankona, line 311: Consider creating a global constant (or ClassVar) for "all", e.g. `INCLUDE_ALL_RELEASE_NAMES="all"`. I see it repeated twice in this method and it might be nice to have more context about what it means (and so we can re-use it on line 328, 364, 380, 395, etc)

roms_tools/setup/cdr_forcing.py
```diff
@@ -503,136 +285,227 @@ def plot_volume_flux(self, start=None, end=None, releases="all"):
...
         data = self.ds["cdr_volume"]
 
         self._plot_line(
             data,
-            releases,
+            release_names,
             start,
             end,
             title="Volume flux of release(s)",
             ylabel=r"m$^3$/s",
```
ankona, line 324: Does ylabel need to be a parameter? I thought we added a "units" field to the dataset previously?

roms_tools/setup/cdr_forcing.py
```diff
@@ -698,82 +573,25 @@ def plot_location_side_view(self, release: str = None):
...
                 "A grid must be provided for plotting. Please pass a valid `Grid` object."
             )
 
-        valid_releases = [r for r in self.releases if r != "_tracer_metadata"]
-        if release is None:
-            if len(valid_releases) == 1:
-                release = valid_releases[0]
+        valid_release_names = [r.name for r in self.releases]
+
+        if release_name is None:
```
ankona, line 586: consider reducing nesting:

```py
if release_name is None and len(valid_release_names) != 1:
  raise ValueError(...)

release_name = valid_release_names[0]
```

roms_tools/setup/cdr_forcing.py
```diff
@@ -873,11 +693,25 @@ def to_yaml(self, filepath: Union[str, Path]) -> None:
             The path to the YAML file where the parameters will be saved.
         """
 
-        _to_yaml(self, filepath)
+        # Serialize object into dictionary
+        yaml_data = _to_dict(self)
```
ankona, line 697: ditto on naming as `yaml_data` when it's a dictionary. you're naming a variable based on who is going to use it, not what it contains.

roms_tools/setup/cdr_forcing.py
```diff
@@ -886,180 +720,254 @@ def from_yaml(cls, filepath: Union[str, Path]) -> "CDRVolumePointSource":
...
-        else:
-            logging.warning(
-                "Grid not provided: cannot verify whether the specified lat/lon/depth location is within the domain or on land. "
-                "Please check manually or provide a grid when instantiating the class."
-            )
 
-    def _validate_release_input(self, releases, list_allowed=True):
-        """Validates the input for release names in plotting methods to ensure they are
-        in an acceptable format and exist within the set of valid releases.
+def _get_release_colors(valid_releases):
```
ankona, line 806: Consider reorganizing plotting methods into a new module to avoid importing unused packages

roms_tools/setup/cdr_forcing.py
```diff
@@ -886,180 +720,254 @@ def from_yaml(cls, filepath: Union[str, Path]) -> "CDRVolumePointSource":
...
+    - Verifies that the location is not below the seafloor.
+    """
+    if grid:
+        # Adjust longitude based on whether it crosses the International Date Line (straddle case)
+        if grid.straddle:
+            lon = xr.where(release.lon > 180, release.lon - 360, release.lon)
+        else:
+            lon = xr.where(release.lon < 0, release.lon + 360, release.lon)
+
+        dx = 1 / grid.ds.pm
```
ankona, line 871: any chance of divide by zero here?

roms_tools/setup/cdr_forcing.py
```diff
@@ -886,180 +720,254 @@ def from_yaml(cls, filepath: Union[str, Path]) -> "CDRVolumePointSource":
...
-        -------
-        dict
-            A dictionary where the keys are release names and the values are their corresponding colors,
-            assigned based on the order of releases in the valid releases list.
 
-        Raises
-        ------
-        ValueError
-            If the number of valid releases exceeds the available colormap capacity.
+def _plot_bathymetry_section(ax, h, dim, fixed_val, coord_deg, resolution, title):
```
ankona, line 919: consider adding type hints

roms_tools/setup/nesting.py
```diff
@@ -209,7 +210,10 @@ def to_yaml(self, filepath: Union[str, Path]) -> None:
             The path to the YAML file where the parameters will be saved.
         """
 
-        _to_yaml(self, filepath)
+        # Serialize object into dictionary
```
ankona, line 213: ditto on naming and extraneous comments

roms_tools/setup/initial_conditions.py
```diff
@@ -926,7 +927,10 @@ def to_yaml(self, filepath: Union[str, Path]) -> None:
             The path to the YAML file where the parameters will be saved.
         """
 
-        _to_yaml(self, filepath)
+        # Serialize object into dictionary
+        yaml_data = _to_dict(self)
```
ankona, line 931: ditto on naming and extraneous comments

roms_tools/setup/grid.py
```diff
@@ -686,29 +687,13 @@ def to_yaml(self, filepath: Union[str, Path]) -> None:
...
-
-        # Use the class name as the top-level key
         yaml_data = {self.__class__.__name__: data}
 
-        with filepath.open("w") as file:
-            # Write header
-            file.write(header)
-            # Write YAML data
-            yaml.dump(yaml_data, file, default_flow_style=False, sort_keys=False)
+        # Write to YAML
```
ankona, line 695: ditto on naming and extraneous comments

roms_tools/setup/cdr_release.py
```diff
@@ -0,0 +1,481 @@
...
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            data = self.model_dump()
+        data["release_type"] = "TracerPerturbation"
+
+        # Flatten tracer_fluxes
+        if "tracer_fluxes" in data:
+            simplified = {}
+            for tracer, contents in data["tracer_fluxes"].items():
```
ankona, line 477: Can we perform this entire action without a loop using the dataset/xarray? This is potentially garbage pseudo-code:

```
tracer_data = self.ds["tracer_fluxes"].loc[all_tracer_names].flatten() # or some stride thing
```

roms_tools/setup/cdr_release.py
```diff
@@ -0,0 +1,481 @@
...
+        if self.times is None:
+            self.times = []
+        num_times = len(self.times)
+
+        if self.tracer_concentrations is None:
+            self.tracer_concentrations = {}
+
+        defaults = get_tracer_defaults()
+        for tracer_name in defaults.keys():
+            if tracer_name not in self.tracer_concentrations:
```
ankona, line 326: consider inverting to reduce nesting...

```py
if tracer_name in self.tracer_concentrations:
  continue

```

or avoid altogether by prefiltering:

```py
filtered_keys = set(defaults.keys()).difference(self.tracer_concentrations)
for tracer_name in filtered_keys:
  ...
```

roms_tools/setup/cdr_release.py
```diff
@@ -0,0 +1,481 @@
...
+
+    def extend_times_to_endpoints(self, start_time, end_time) -> None:
+        """Ensures that `times` includes both `start_time` and `end_time`.
+
+        Modifies `self.times` in place by prepending or appending times as needed.
+        If `times` is empty, it will be set to [`start_time`, `end_time`].
+        """
+
+        if not self.times:
+            self.times = [start_time, end_time]
```
ankona, line 239: It seems like this base case should not be part of this method. Initializing `self.times` is not the same as extending, so we're doing 2 things in one place. I have a feeling that you really want something in a `post_init` method to initialize `self.times`

roms_tools/setup/cdr_release.py
```diff
@@ -0,0 +1,481 @@
...
+        time window.
+
+        Raises
+        ------
+        ValueError
+            If times are not strictly increasing, or fall outside the [start_time, end_time] window.
+        """
+
+        if self.times and len(self.times) > 0:
+            if not all(t1 < t2 for t1, t2 in zip(self.times, self.times[1:])):
```
ankona, line 225: consider `np.diff(self.times)` instead of the zipping trick. you can do

```
if np.any(np.diff(self.times) <= 0):
  ...
```

roms_tools/setup/cdr_release.py
```diff
@@ -0,0 +1,481 @@
...
+            Start of the interval to cover.
+        end_time : datetime-like
+            End of the interval to cover.
+
+        Returns
+        -------
+        self : Concentration
+            Updated instance with extended values.
+        """
+        start_pad = self.values[0] if isinstance(self.values, list) else self.values
```
ankona, line 174: consider a single conditional:

```py
if isinstance(self.values, list):
  start_pad = self.values[0]
  end_pad = self.values[-1]
else:
  start_pad = self.values
  end_pad = self.values
```

And, now that I see that... isn't there a bug in `end_pad` being the same as `start_pad` when it isn't a list?

roms_tools/setup/cdr_release.py
```diff
@@ -0,0 +1,481 @@
...
+    name : str
+        Name of the variable (e.g., flux or concentration).
+    values : Union[float, List[float]]
+        A constant value or a time-varying series of values.
+    """
+
+    name: str
+    values: Union[float, List[float]]
+
+    def check_length(self, num_times: int):
```
ankona, line 29: -> None

roms_tools/setup/river_forcing.py
NoraLoose: > alternatively, consider NOT formatting here and instead performing formatting in a single place (the yaml serializer). so, here, we would ideally end up only creating the data to be serialized. The strength of this approach is that this method doesn't produce values that are tightly coupled to the consumer code.

Are you suggesting that the custom formatting could go in `_to_yaml` and `_from_yaml`? In this PR, I actually pulled those formatting rules *out* of `_to_yaml` / `_to_dict` on purpose. They're specific to `RiverForcing`, and nothing else uses them.

I like keeping the formatting and reverse logic in the same place, so it’s easy to follow. For example:

- [Formatting logic](https://github.com/NoraLoose/roms-tools/blob/0af519ed13ac895df101476a5e05b1c58fef584e/roms_tools/setup/river_forcing.py#L680-L687)  
- [Reverse logic](https://github.com/NoraLoose/roms-tools/blob/0af519ed13ac895df101476a5e05b1c58fef584e/roms_tools/setup/river_forcing.py#L710-L721)

Keeps it all in one spot without cluttering the general YAML stuff.

Does this make sense to you?

NoraLoose: I agree with this, though:

> Feel free to get rid of the comment stating "comma-separated string."

Done in [d94e9a1](https://github.com/CWorthy-ocean/roms-tools/pull/301/commits/d94e9a10d3328f33cef85e88e187d00cc40926e4)

roms_tools/setup/river_forcing.py
NoraLoose: Done in af3d3bc.

roms_tools/setup/cdr_forcing.py
NoraLoose: Good point and great suggestion. Thanks! Changed in [49ef5cb](https://github.com/CWorthy-ocean/roms-tools/pull/301/commits/49ef5cbacacd4311e673856c7c044dbfe4c7e68b).

roms_tools/setup/cdr_forcing.py
NoraLoose: Nice. Done in a02135f.

roms_tools/setup/cdr_forcing.py
NoraLoose: Good point. Done in 44ed95d.

roms_tools/setup/cdr_forcing.py
NoraLoose: Sounds good. Done in a5f12ae (except that it's  a method, not a classmethod).

NoraLoose:
@ScottEilerman the tests with Python 3.10 are failing because `StrEnum` was added in Python 3.11. Should we simply require Python >= 3.11?

ScottEilerman:
> @ScottEilerman the tests with Python 3.10 are failing because `StrEnum` was added in Python 3.11. Should we simply require Python >= 3.11?

[3.11 has been out long enough](https://devguide.python.org/versions/) that I don't feel bad requiring it. If there's any concern though, I can make a small commit to change it to a regular enum.

NoraLoose:
> [3.11 has been out long enough](https://devguide.python.org/versions/) that I don't feel bad requiring it. If there's any concern though, I can make a small commit to change it to a regular enum.

I'm okay with dropping support for Python 3.10. Will speed up running the CI as well! :)

roms_tools/setup/cdr_forcing.py
ScottEilerman: FWIW I prefer the newer version and changed a few of them. Though I don't have strong feelings if others prefer the Old Ways.

NoraLoose: We want the new style! Thanks!

roms_tools/setup/cdr_forcing.py
NoraLoose: The `ylabel` here has nicer LaTeX formatting.

roms_tools/setup/cdr_forcing.py
NoraLoose: This won't work because it always overwrites `release_name`, even when the user has already specified it (i.e., it's not None).

roms_tools/setup/cdr_forcing.py
NoraLoose: We should do the same thing for all other roms-tools classes. Maybe in a future PR?

roms_tools/setup/cdr_forcing.py
NoraLoose: No... `grid.ds.pm` is always bigger than zero!

roms_tools/setup/cdr_forcing.py
NoraLoose: Done in 8fa5360.

roms_tools/setup/cdr_release.py
NoraLoose: The Release objects don't have access to the xarray Dataset.

roms_tools/setup/cdr_release.py
NoraLoose: done in d2aeeaf

roms_tools/setup/cdr_release.py
NoraLoose: done in [849833e](https://github.com/CWorthy-ocean/roms-tools/pull/301/commits/849833e8291e7c1a17c3246129fbaf71e1f34fcd)

roms_tools/setup/cdr_release.py
NoraLoose: > And, now that I see that... isn't there a bug in end_pad being the same as start_pad when it isn't a list?

That's intentional. If it isn't a list, it just uses the given value as a constant in time

NoraLoose: Using a single conditional as of [8b2fba4](https://github.com/CWorthy-ocean/roms-tools/pull/301/commits/8b2fba44a9ad35b256def217c575c17403ead6fd)

roms_tools/setup/cdr_release.py
NoraLoose: We cannot initialize `self.times` with `[start_time, end_time]` in `post_init` because subsequent checks verify that the number of times matches the number of provided fluxes and concentrations. Initializing early would cause these checks to fail due to a mismatch.

roms_tools/setup/river_forcing.py
NoraLoose: Back to the `to_yaml` / `to_dict` question: I think we can solve this easily if we switch the `RiverForcing` class to pydantic in the future. Let's leave this for a future PR?!

roms_tools/setup/cdr_forcing.py
NoraLoose: Done in [ed368b1](https://github.com/CWorthy-ocean/roms-tools/pull/301/commits/ed368b16584b9ff66b3fba2ae92ad620d414e7ae).

PR #315, opened 2025-05-23
Enable and run isort

ScottEilerman:
This PR is formatting-only shouldn't affect code. It enables import sorting via ruff's isort plugin. It looks like isort had some boilerplate configuration in the pyproject.toml already, but was not actually being called in pre-commit. I added the ruff hook in pyproject.toml, removed the unneeded isort config, and ran `pre-commit run --all-files`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,26 +1,28 @@
-import pytest
+import hashlib
 from datetime import datetime
+from pathlib import Path
+
+import h5py
+import pytest
+
 from roms_tools import (
-    Grid,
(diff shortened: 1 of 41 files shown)


NoraLoose:
That's a great addition - thanks @ScottEilerman! The unsorted imports have always been bothering me, but not enough to do anything about it. :)

NoraLoose (approved):

NoraLoose:
Feel free to merge!

Issue #293, opened 2025-04-10
Test suite `CI / 3.13-build-with-dask-and-xesmf (3.13)` hanging


NoraLoose:
### Summary

The CI job `CI / 3.13-build-with-dask-and-xesmf (3.13)` hangs indefinitely, preventing the test matrix from completing. If a PR is merged before the job finishes (which it never will), the `main` branch eventually marks the job as failed due to timeout, causing CI to report a failure.

---

### ✅ What Works
- `3.13` with only `dask` runs successfully.

### ❌ What Fails
- `3.13` with both `dask` and `xesmf` hangs.

ankona:
Did you try adding debug logging so pytest will print out verbose logs? If not, it might give us the details on some failure. 

Try adding the following to `pyproject.toml` in your branch:

```ini
[tool.pytest.ini_options]
cli_log_level=DEBUG
```

Alternatively, it might be better to parameterize the GH build action. Use an env var to set the log level and modify the call triggering the tests. Modify call to pytest in `unit_tests.yaml` to:

```
# see: https://docs.pytest.org/en/stable/how-to/output.html#verbosity
name: unit_tests
on:
 ...
env:
  # can be set in github repo admin instead of hardcoded 
  CLI_LOG_LEVEL: DEBUG
...

 jobs:
  steps:
    ... # see: https://github.com/cclauss/GitHub-Action-for-pytest?tab=readme-ov-file#github-action-for-running-pytest-commands
    ... pytest --log-cli-level=$CLI_LOG_LEVEL
```

I think this might get you the logs on where the thing is hanging.

NoraLoose:
We may have fixed this by PR #501

PR #327, opened 2025-06-02
Add tracer metadata method and include tracer units in YAML exports

### LINKED ISSUE #305: Add utility function to return expected tracer units for `VolumeRelease` and `TracerPerturbation`
As a follow-up to #301, we should provide a utility function that returns the **expected units** for:

- **Tracer concentrations** (required when defining a `VolumeRelease`)
- **Tracer fluxes** (required when defining a `TracerPerturbation`)

Currently, these units appear in the `.ds` attribute, but only **after** the user has constructed `Release` objects and passed them to `CDRForcing`. This makes it hard for users to know in advance what units to use when specifying their input.

### LINKED ISSUE #307: Do we want tracer metadata in the `CDRForcing` YAML?
Before #301, the YAML files included full tracer metadata. However, this made the YAML files very long and harder to read.  
For example, the screenshot below shows only the first 10 out of 34 entries:

Old metadata format

After #301, we stopped including this metadata in the YAML altogether.

### Question

Do we want to bring the tracer metadata back into the YAML?

If yes, we might also consider a more compact or readable format. For example, instead of placing all metadata at the top or bottom of the releases section within the YAML file (disconnected from each individual release), we could include relevant unit information **inline with each release**, which could make the YAML more readable.

Inline metadata idea

Happy to hear thoughts!

NoraLoose:
This PR improves transparency around tracer units for users when constructing their CDR Forcing by:

* Adding a `.get_tracer_metadata()` method that can be called without instantiating a release. This allows users to inspect expected tracer names and units ahead of time. See #305.
* Writing the tracer metadata (`_tracer_metadata`) into the YAML file when saving. See #307.

- [x] Closes #305, closes #307.
- [x] Tests added


NoraLoose:
The 2 failing checks were cancelled after hours. I think this is issue #293 again. Anyway, the CDR forcing module does not use dask, and the xesmf tests are passing locally for me.

roms_tools/setup/cdr_forcing.py
```diff
@@ -789,10 +789,16 @@ def to_yaml(self, filepath: str | Path) -> None:
...
-        # Serialize object into dictionary
         forcing_dict = self.model_dump()
 
-        # Write to YAML
+        if self.release_type == ReleaseType.volume:
+            metadata = VolumeRelease.get_tracer_metadata()
+        elif self.release_type == ReleaseType.tracer_perturbation:
+            metadata = TracerPerturbation.get_tracer_metadata()
+        else:
+            metadata = {}
```
ScottEilerman, line 799: I guess since we already guarantee that there is at least one release and all releases of the same type, we could reduce these if/else blocks to

```python
metadata = self.releases[0].get_tracer_metadata()
```

right? Maybe a little cleaner, and we get the behavior for any new Release classes without needing to revisit this block in the future.

Although that implies we should add a `get_tracer_metadata()` to the base `Release` that either:
* raises a NotImplementedError
* returns {}
* is marked as an abstractmethod, and make Release an ABC (though that will break some tests I think)

ScottEilerman (approved):
One suggestion, but ok as-is if you want to move on.

PR #345, opened 2025-06-13
Visualize Gaussian distribution associated with CDR releases

### LINKED ISSUE #306: Enhance `CDRForcing` location plots to show Gaussian shapes
As a follow-up to PR #301, we should improve the plotting methods of `CDRForcing`:

- `.plot_location_top_view()`
- `.plot_location_side_view()`

### Current Behavior

Both methods currently show only the center of each release using a cross (`x`). This works fine when the release is effectively a point source.

### Desired Enhancement

If the user specifies `hsc > 0` (horizontal standard deviation) or `vsc > 0` (vertical standard deviation), the plots should display a **Gaussian distribution shape** instead of just a single point. The Gaussian shape plotted should (more or less) **mimic** how ROMS internally distributes tracers using `hsc` and `vsc`.

NoraLoose:
This PR adds the method `CDRForcing.plot_distribution()`, which visualizes the Gaussian distribution associated with each release. The plotting mimics how the ROMS CDR module maps the Gaussian distribution in both horizontal and vertical directions, replicating its behavior as closely as possible for visualization purposes.

### Example output
Screenshot 2025-07-01 at 12 56 39 PM

### Internal changes
* Moved `_compute_exponential_depth_levels` from `ROMSOutput` to `Grid` (where it logically belongs). Although this method was initially used for this feature, it is not used in the final implementation.
*Added `_generate_focused_coordinate_range`, which builds a focused coordinate grid (for plotting only) with higher resolution near the release center.
* Fixed a bug in `_remove_edge_nans` so it works when `depth_layers` is missing or equal to topography. This fix is used by the new plotting functionality.

### Checklist
- [x] Closes #306
- [x] Tests added


NoraLoose:
<p>Thanks for catching this.</p>

---

 <a href='https://app.reviewnb.com/CWorthy-ocean/roms-tools/pull/345/discussion/#comment-2985641407'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

roms_tools/setup/cdr_forcing.py
```diff
@@ -986,66 +990,298 @@ def _validate_release_location(grid, release: Release):
...
+        Normalized 2D tracer distribution on the ROMS grid (zero over land).
+    """
+    # Find closest grid cell center
+    target_coords = get_target_coords(grid)
+    lon = release.lon
+    if target_coords["straddle"]:
+        lon = xr.where(lon > 180, lon - 360, lon)
+    else:
+        lon = xr.where(lon < 0, lon + 360, lon)
+    dist = gc_dist(target_coords["lon"], target_coords["lat"], lon, release.lat)
```
NoraLoose, line 1019: @ScottEilerman Here we want to use your new numba function I think!

NoraLoose: Oh, but we don't want minimum distance to land, but just the grid indices with minimum distance.

ScottEilerman: I actually had a version where there was a more general "min dist between one array of points and another array of points" but I liked the more specific version better and figure we can back out the more general one later if we want it. But gc_dist should be vectorized and broadcast fine too.

ScottEilerman (approved):

PR #353, opened 2025-06-18
Report topography source path in NetCDF grid file

### LINKED ISSUE #349: Grid.save() doesn't add path for topography_source
After making a `Grid` object, `grid`, and saving it as a netCDF (`grid.save(filepath)`) there is no path stored for `topography_source`. This is a problem when later loading a netCDF file to use when making new objects (`BoundaryForcing` for example). Ultimately this results in an error when trying to load a `BoundaryForcing` from a yaml, which results in this error:

```
ValueError: `topography_source` must include a 'path' key when the 'name' is not 'ETOPO5'.
```

Code to reproduce a yaml without the topo src path:
```
from roms_tools import Grid
from datetime import datetime

grid = Grid(
    nx=500,
    ny=500,
    size_x=1000,
    size_y=1000,
    center_lon=-37,
    center_lat=64,
    rot=25,
    N=100,
    topography_source={
        "name": "SRTM15",
        "path": "/global/cfs/projectdirs/m4746/Datasets/SRTM15/SRTM15_V2.6.nc",
    },  
)

filepath = "./test_grid.nc"
grid.save(filepath)


grid2 = Grid.from_file(filepath)
print(grid2)
```

NoraLoose:
Before this PR, all NetCDF files included only the name from `source_data = {"name": name, "path": path}` as an attribute. This minimal metadata approach was intentional, since reproducibility is primarily ensured through the YAML configuration files.

However, the grid NetCDF file requires special treatment: users can recreate the grid object without a YAML file via `Grid.from_file(filepath)`. In this case, it is essential to record both the name and the full path of the topography source to ensure traceability.

This PR updates the grid NetCDF behavior as follows:
* Added attributes:
  * `topography_source_name`
  * `topography_source_path`
* Removed attributes:
  * `topography_source`

I have also moved all topography-related tests from the `test_topography.py` module to the `test_grid.py` module. This allows the existing grid fixtures that use SRTM15 topography data to be reused for testing roundtrip behavior with the new `topography_source_name` and `topography_source_path` attributes.

- [ ] Closes #349
- [ ] Tests added


NoraLoose:
> I don't get an error when partitioning

Just to clarify. Did this issue contribute to partitioning errors? I'm confused now.

PR #355, opened 2025-06-18
Stream era5

ScottEilerman:
Really the only significant thing that needs to happen to make this thing work is using the different variable names, and reading zarr files with different params to open_dataset or open_zarr (no combine keywords, and can't use mfdataset).

Dataset info for reference: https://github.com/google-research/arco-era5

TODOs:
- investigate/fix broken test
- refactor as desired (especially w.r.t. ERA5_ARCO naming)
- probably add more tests
- docs

- [ ] Closes #xxxx
- [ ] Tests added


roms_tools/utils.py
```diff
@@ -151,14 +162,22 @@ def _load_data(
...
+                )
+            else:
+                ds = xr.open_mfdataset(
+                    matching_files,
+                    decode_times=decode_times,
+                    decode_timedelta=decode_times,
+                    chunks=chunks,
+                    **combine_kwargs,
+                    **kwargs,
+                )
```
NoraLoose, line 180: Can this be consolidated to
```python
                ds = xr.open_mfdataset(
                    matching_files,
                    engine=engine,
                    decode_times=decode_times,
                    decode_timedelta=decode_times,
                    chunks=chunks,
                    **combine_kwargs,
                    **kwargs,
                )
```
where `engine` is "netcdf4" or "zarr"?

NoraLoose:
@ScottEilerman thanks for working on this. I will try to make some more progress on this PR on Friday, once Perlmutter is up again.

roms_tools/utils.py
ScottEilerman: I don't think we can use `mfdataset` for zarr. `open_dataset(..., engine="zarr")` works instead of `open_zarr`, but neither will allow some of the parameters we put together for the `open_mfdataset` call.

PR #356, opened 2025-06-18
Handling overlapping rivers

### LINKED ISSUE #260: Non-Unique `river_location` When Rivers Overlap
### Issue

When two rivers overlap on the same grid point, the current ROMS-internal rule for computing `river_location`:  
```
river_location = river ID + fraction of the river's flux through the grid point.
```
is no longer unique (as we will see in more detail in the simplified example below). The situation of overlapping rivers can arise when the grid resolution is insufficient, causing rivers to be moved to the same location by an automated procedure. This can be seen in this figure, where `ROMS-Tools` moved the purple river onto the orange river in order to satisfy the criterion that each river has to lie on the coast.

Image

Of course we could avoid moving two rivers to the same location, but I think this would just be a work-around for a bigger underlying issue.

### Simplified Example

Consider two rivers:

* River 1 spans grid points [(1, 2), (1, 3)]
* River 2 spans grid points [(1, 3), (2, 3)]

Both rivers contribute to the grid point (1,3), leading to conflicting assignments:

```
# River 1 contribution
river_location = 1 + 0.5  # 1.5

# River 2 contribution
river_location = 2 + 0.5  # 2.5
```
If these values are summed, the result is:
```
river_location = 1.5 + 2.5  # 4.0
```
This mistakenly implies that a third river (ID 3) is occupying grid point (1,3) with fraction 1.0, causing an incorrect interpretation of river locations.

### Discussion

How can we ensure unique representation of river location and fraction? 

A simple fix that we have discussed  is to separate `river_location` into two variables:
```
river_ID[eta_rho, xi_rho]
fraction[eta_rho, xi_rho]
```
However, this alone does not resolve overlapping rivers because multiple rivers can still map to the same `(eta_rho, xi_rho)`, making it unclear which fraction belongs to which river.

### Suggested solution

Instead, I propose to introduce a single variable
```
fraction[eta_rho, xi_rho, nriver]
```
where `nriver` corresponds to the river ID. This solution would ensure that each river’s fraction is stored separately without interference.

What do you think @ScottEilerman @dafyddstephenson? This is relevant for UCLA-ROMS development upstream.

NoraLoose:
This PR introduces the method `.handle_overlapping_rivers`, which is called as part of creating the river forcing. 

`.handle_overlapping_rivers` works as follows: 

If multiple rivers are assigned to the same grid cell (i.e., overlapping index pairs), this method creates new uniquely named rivers ('overlap_1', 'overlap_2', ...) that consolidate the contributions from all original rivers sharing that location.

For each overlapping grid cell:
* A new river is created using a volume-weighted sum of the original rivers' volume and tracer.
* The volume fraction of the original river is reduced proportionally to reflect the removal of the shared grid cell.
* The new river is appended to the dataset and indexed in `self.indices`.
        
- [x] Closes #260
- [x] Tests added


roms_tools/setup/river_forcing.py
```diff
@@ -328,80 +405,129 @@ def _create_river_forcing(self, data):
...
+
+            for i, (idx_pair, river_list) in enumerate(overlapping_rivers.items()):
+                new_name = f"overlap_{i+1}"
+                self.indices[new_name] = [idx_pair]
+
+                # Get index of all rivers contributing to this overlapping cell
+                contributing_indices = [
+                    np.where(ds["river_name"].values == name)[0].item()
+                    for name in river_list
+                ]
```
ScottEilerman, line 461: My xarray is rusty. Can we simplify this to something like `ds.sel(river_name=river_list)`?

roms_tools/setup/river_forcing.py
```diff
@@ -328,80 +405,129 @@ def _create_river_forcing(self, data):
...
+                            "float64"
+                        )
+                        for i, weight in zip(contributing_indices, weighted_volumes)
+                    ],
+                    dim="tmp",
+                )
+                combined_river_tracer = (
+                    weighted_tracers.sum(dim="tmp", skipna=True) / combined_river_volume
+                )
+                # If combined_river_volume is 0.0, the result will be NaN. Replace with default for clarity.
```
ScottEilerman, line 489: Why would the combined river volume ever be zero?

roms_tools/setup/river_forcing.py
```diff
@@ -328,80 +405,129 @@ def _create_river_forcing(self, data):
...
 
-        return river_indices
+                combined_river_volumes.append(combined_river_volume)
+                combined_river_tracers.append(combined_river_tracer)
+
+                # Reduce volume fraction of each original river by appropriate amount
+                for i, n_cells in zip(contributing_indices, num_cells_per_river):
+                    ds["river_volume"].loc[{"nriver": ds.nriver[i]}] = (
+                        ds["river_volume"].isel(nriver=i) * (n_cells - 1) / n_cells
+                    )
```
ScottEilerman, line 517: The body of this for loop could be a good exercise in single responsibility principle. I thought this last bit about reducing the volume of the original rivers was going to make it complicated, but actually it can be independent too. I think you could break the bulk of it into a `new_river_datarray = self._create_combined_river(ds, idx_pair, river_list)` and then this last bit into `ds = self._reduce_river_volume(ds, contributing_indices, num_cells_per_river)`. Those smaller functions are then more easily and independently testable, and this `_handle_overlapping_rivers` method becomes a lot shorter and clearer.

ScottEilerman (approved):
A few suggestions, but also ok as-is if you want to move on.

roms_tools/setup/river_forcing.py
```diff
@@ -233,6 +235,81 @@ def _get_data(self):
...
+        dist_coast = gc_dist(
+            target_coords["lon"].where(coast),
+            target_coords["lat"].where(coast),
+            river_lon,
+            river_lat,
+        ).transpose(data.dim_names["station"], "eta_rho", "xi_rho")
+        dist_coast_min = dist_coast.min(dim=["eta_rho", "xi_rho"])
+
+        # Find the indices of the closest coastal grid cell to the river mouth
+        indices = np.where(dist_coast == dist_coast_min)
```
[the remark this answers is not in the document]
NoraLoose: > actually, thinking a little more, the existing one might work if you give it the coast mask as the "land" mask. Though the function and variable naming is still awkward then.

I think `min_dist_to_land` would only give you the actual minimal distance, but not the **index** of the closest coastal grid cell, correct?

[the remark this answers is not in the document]
ScottEilerman: Correct, you'd still have to do the `where` lookup if we went that route.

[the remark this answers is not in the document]
NoraLoose: Cool. Let's use xarray's `argmin` for now. I implemented this in a new function in [e98ca98](https://github.com/CWorthy-ocean/roms-tools/pull/356/commits/e98ca984d723e9efdde84e0660c63e81ae8ea1a9). This function is then used twice within the river forcing dataset generation.

roms_tools/setup/river_forcing.py
NoraLoose: Good point. I attempted this via the past three commits ([d195851](https://github.com/CWorthy-ocean/roms-tools/pull/356/commits/d195851bfde4a5973b6f38912b4d460d3d9ad53a), [2251ab7](https://github.com/CWorthy-ocean/roms-tools/pull/356/commits/2251ab7a978c77726e49e10900118f9babbc74f1), [bd03351](https://github.com/CWorthy-ocean/roms-tools/pull/356/commits/bd03351fb79de03c50377cc38047b060de3aad38))

PR #357, opened 2025-06-23
Stream ARCO ERA5 data

NoraLoose:
This PR is in collaboration with @ScottEilerman. I started from #355, but somehow could not push to it directly. The PR implements streaming ERA5 data directly from the cloud so that users do not need to pre-download their ERA5 data.

### Changes

- Introduced `ERA5ARCODataset` class.
- Added `gcsfs` as an optional dependency to support cloud streaming.  
- Introduced a new pytest marker `stream` to isolate streaming tests:  
```bash
  pytest -m stream --stream
```
This runs only streaming tests and excludes them otherwise. This allows faster development since streaming tests take longer.

- Added streaming tests to CI.
- Added ERA5 streaming to documentation.

Other changes:

- Reorganized internal logic for handling `use_dask` and `read_zarr` workflows.  
- Cleaned up `conftest.py` by removing unnecessary `request` parameters from fixtures.  
- Adjusted fixture `start_time` values in `conftest.py` to more intuitive dates (changed from January 31 to February 2) since source data does not include January.  

- [x] Tests added


NoraLoose:
@ScottEilerman I'm still getting the warning:
```
RequestsDependencyWarning: Unable to find acceptable character detection dependency (chardet or charset_normalizer).
  warnings.warn(
```
in the docs notebook when instantiating the `SurfaceForcing` object with streaming. We can leave it for a future PR to try to fix that.

Issue #364, opened 2025-06-30
Variable name changed?


smaticka:
In the **Visualizing ROMS Output** section in the readthedocs [here](https://github.com/CWorthy-ocean/roms-tools/blob/23e61486cd5af1154db31119c8d6ed47fcabffa3/docs/plotting_roms_output.ipynb), the plotting functionality is shown with `roms_output.plot("MARBL_PH_3D", time=0, s=-1)`. When I ran it on my `*cstar.nc` output file, I didn't have that variable, though I did have the next `DIC` variable. 

Should this be changed to `pH`? or should we add a single line description on the `roms_output.plot` usage? I don't think a description is needed if the first example uses a variable that will always be in the output file. (e.g. not `ALK` in case they didn't do a CDR release)

NoraLoose:
You can only plot the variables that are present in the output. The variables that are present can be inspected via `roms_output.ds`.

`MARBL_PH_3D` is a variable that is present in the restart files, but it is not present in the C-Star output.

NoraLoose:
You are welcome to open a PR that clarifies the above! :)

smaticka:
Thanks. Maybe it is clear enough to other readers, but it took me a minute to figure out. I'll add a one-liner PR - I figure it won't hurt to have extra clarity :)

PR #366, opened 2025-06-30
Adding short helper to find ROMS output

smaticka:
Adding a sentence to the notebook for `Visualizing ROMS output`, just to make it a little easier to understand what variables are able to be plotted. 
It's particularly helpful if the reader is using a different output file type than the one in the example given (avg vs rst for example), and they aren't able to plot the example variables. 

Resolves: #364

### DIFF
--- a/docs/plotting_roms_output.ipynb
+++ b/docs/plotting_roms_output.ipynb
@@ -115,7 +115,7 @@
    "id": "a8266ffc-ec66-4859-8480-1f61d985d930",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


docs/plotting_roms_output.ipynb
```diff
@@ -112,10 +112,18 @@
   },
```
NoraLoose, line 112: <p>Final nit: Should we add a period after the sentence?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CWorthy-ocean/roms-tools/pull/366/discussion/#comment-2178539449'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

PR #381, opened 2025-07-18
Better match grid latitudes/longitudes and bathymetry from UCLA tools

NoraLoose:
The goal of this PR is to improve consistency with the UCLA tools by more closely matching the horizontal grid (lat/lon coordinates) and bathymetry. Specifically, it updates:

* Latitude/longitude generation: Coordinates are now computed to match those from the UCLA tools to within ~10⁻⁴ degrees.

* Bathymetry smoothing:
  * The domain-wide smoothing scale is reduced (this is the key change).
  * An area-weighted smoothing scheme is used (though its effect is minor when grid cells are of similar size).

### Before this PR
Screenshot 2025-07-18 at 2 58 00 PM

Screenshot 2025-07-18 at 2 57 46 PM

### After this PR

Screenshot 2025-07-18 at 2 53 36 PM
Screenshot 2025-07-18 at 2 53 48 PM

## Bathymetry example: Off the American East Coast

Plotting the Laplacian of the bathymetry is a good way to see the smoothness of the field. 

### Before adjusting the smoothing scale

Screenshot 2025-07-21 at 1 30 15 PM

### After this PR

Screenshot 2025-07-21 at 1 31 24 PM


roms_tools/setup/grid.py
```diff
@@ -987,38 +987,58 @@ def _make_initial_lon_lat_ds(self):
...
+            dlat /= mul
+
+        # Latitudes of corners
         latq_array_1d_in_degrees = np.arctan(
-            np.sinh((y2 - y1) * (np.arange(-1, nw + 2) / nw) + y1)
+            np.sinh((y2 - y1) * np.arange(-1, nw + 2) / nw + y1)
         )
 
+        # Rescale
+        latq_array_1d_in_degrees = latq_array_1d_in_degrees / mul
```
ScottEilerman, line 1037: Is any of the rest of this supposed to be in the loop? Double-checking since some of the variables that get changed in the loop (`mul`, `y1`, `y2`, etc.) also get used down here.

NoraLoose: I was wondering the same thing. But that's not what they do here: https://github.com/nmolem/ucla-tools/blob/main/EGRID_EXP/easy_grid.m

I think there are a few bugs/inconsistencies in the UCLA code. The question is: Do we want to reproduce as closely as possible or do we want to have some half-way reasonable code?

NoraLoose: The most consistent thing would be at least to rescale `lat_array_1d_in_degrees` once more (next to `latq_array_1d_in_degrees`) after the loop.

NoraLoose:
@ScottEilerman after [75de0d1](https://github.com/CWorthy-ocean/roms-tools/pull/381/commits/75de0d127bfcce1b75c79c3f28d5550a9086d9e1), the comparison figures in the PR description did not change.

PR #394, opened 2025-07-30
Make `partition_netcdf` more flexible

### LINKED ISSUE #259: `partition_netcdf` should accept `List[Path]`
as `ROMSToolsObject.save` no longer accepts `np_xi`, `np_eta` parameters, we could streamline partitioning by supporting patterns like 

```
output_path_list = roms_forcing.save("my_forcing_files")
roms_tools.partition_netcdf(output_path_list, np_xi=np_xi, np_eta=np_eta)
```
such that the list of paths returned by `save` is accepted (and iterated over) by `partition_netcdf`. This would also allow batch partitioning (as all files generated in a `roms-tools` session are ultimately likely to be partitioned in the same way).

### LINKED ISSUE #350: Option to specify directory for partition files. nice2have
Currently the behavior of `partition_netcdf` takes a file path to the location of the NetCDF file to be partitioned, and `partition_netcdf` creates a number of partitioned files and stores them in the same directory where the file path points. 
This creates a lot of files in a single directory and it can be difficult to find the original files. 

It would be nice to be able to optionally specify a directory to tell `partition_netcdf` where to store the partitioned files. 

Current functionality of `partition_netcdf`:
`partition_netcdf(filepath, np_eta=5, np_xi=5)`

### LINKED ISSUE #388: Provide output directory for partitioned files
When using the partitioning function, the partitioned files end up in the same directory as the un-partitioned files. Can we add some functionality so that you can output the partitioned files into a specified directory? Otherwise, we need to constantly move all the partitioned files afterward which can become very cumbersome.

NoraLoose:
- [ ] Closes #388, closes #350, closes #259
- [ ] Tests added

### DIFF
--- a/roms_tools/tiling/partition.py
+++ b/roms_tools/tiling/partition.py
@@ -299,6 +299,7 @@ def partition_netcdf(
     filepath: str | Path,
     np_eta: int = 1,
     np_xi: int = 1,
+    output_dir: str | Path | None = None,
     include_coarse_dims: bool = True,
 ) -> None:
     """Partition a ROMS NetCDF file into smaller spatial tiles and save them to disk.
@@ -316,6 +317,10 @@ def partition_netcdf(
 
     np_xi : int, optional
         The number of partitions along the `xi` direction. Must be a positive integer. Default is 1.
+
+    output_dir : str | Path | None, optional
+        Directory or base path to save partitioned files.
+        If None, files are saved alongside the input file.
 
     include_coarse_dims : bool, optional
         Whether to include coarse grid dimensions (`eta_coarse`, `xi_coarse`) in the partitioning.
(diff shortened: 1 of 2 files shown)


roms_tools/tiling/partition.py
```diff
@@ -296,59 +297,72 @@ def cumsum(pmf):
 
 
 def partition_netcdf(
-    filepath: str | Path,
+    filepath: str | Path | Sequence[str | Path],
     np_eta: int = 1,
     np_xi: int = 1,
+    output_dir: str | Path | None = None,
     include_coarse_dims: bool = True,
 ) -> None:
```
ankona, line 305: typehint mismatch to actual return value

docs/partition.ipynb
```diff
@@ -438,235 +438,237 @@
        "Dimensions:       (eta_rho: 152, xi_rho: 302, xi_u: 301, eta_v: 151,\n",
```
dafyddstephenson, line 438: <p>Think this is not the error you wanted to display! </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CWorthy-ocean/roms-tools/pull/394/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

dafyddstephenson (approved):
Small issue in one of the notebook cells and maybe should open a ticket about including MyPy but looks good!

roms_tools/tiling/partition.py
NoraLoose: > This should have been caught by pre-commit but I'm just noticing there's no MyPy on pre-commit?

I opened https://github.com/CWorthy-ocean/roms-tools/pull/405

PR #405, opened 2025-08-06
Add mypy to pre-commit

NoraLoose:


roms_tools/setup/grid.py
```diff
@@ -758,24 +759,24 @@ def from_file(
...
                 }
             else:
-                a = {"name": ds.attrs["topography_source_name"]}
+                topo_source = {"name": ds.attrs["topography_source_name"]}
         else:
-            a = None
+            topo_source = None
 
-        object.__setattr__(grid, "topography_source", a)
+        object.__setattr__(grid, "topography_source", topo_source)
```
ankona, line 779: `topography_source` appears to be a pre-defined attribute of the `Grid` base class. It should be safer if the typechecker can validate the assignment using `grid.topography_source = a`. 

Is there a reason to use `__setattr__`?

roms_tools/setup/grid.py
```diff
@@ -881,15 +885,15 @@ def __repr__(self) -> str:
...
+    def _create_horizontal_grid(self) -> xr.Dataset:
         """Create the horizontal grid based on a Mercator projection and store it in the
         'ds' attribute.
 
         Parameters
         ----------
         None
 
-        Returns
+        Returns:
```
ankona, line 896: I think we have some errant items, but we're on the "no colon" side. Did you mean to make the separator `-------` a single char shorter instead?

roms_tools/setup/nesting.py
```diff
@@ -77,11 +82,17 @@ class ChildGrid(Grid):
 
     def __post_init__(self):
         super().__post_init__()
-        self._map_child_boundaries_onto_parent_grid_indices()
-        self._modify_child_topography_and_mask()
+        self._map_child_boundaries_onto_parent_grid_indices(verbose=self.verbose)
+        self._modify_child_topography_and_mask(verbose=self.verbose)
```
ankona, line 86: it doesn't make a lot of sense to have `self.verbose` be a member variable and then pass it as a parameter - isn't it already available?

roms_tools/setup/nesting.py
```diff
@@ -94,14 +105,24 @@ def _map_child_boundaries_onto_parent_grid_indices(self):
             self.metadata["period"],
         )
 
+        if verbose:
+            logging.info(f"Total time: {time.time() - start_time:.3f} seconds")
+            logging.info(
+                "========================================================================================================"
```
ankona, line 111: Consider using another constant or even an int to avoid having these long, dupe lines.

```python
HEADER_SZ = 80
HEADER_CHAR = "="

@classmethod
def _log_header(cls) -> str:
    return cls.HEADER_CHAR * cls.HEADER_SZ
```

or 

```python
@classmethod
def log_the_separator(cls) -> None:
    logging.info(cls.HEADER_CHAR * cls.HEADER_SZ)
```

roms_tools/setup/nesting.py
```diff
@@ -269,13 +304,28 @@ def _finalize_grid_datasets(
...
-    def from_file(cls, filepath: str | Path, verbose: bool = False) -> "ChildGrid":
+    def from_file(
+        cls,
+        filepath: str | Path,
+        theta_s: float | None = None,
+        theta_b: float | None = None,
+        hc: float | None = None,
+        N: int | None = None,
+        verbose: bool = False,
+    ) -> "ChildGrid":
```
ankona, line 328: Consider expanding the name of `N` so it's purpose is clear (and so it meets naming standards).

roms_tools/setup/river_forcing.py
```diff
@@ -691,7 +703,11 @@ def plot_locations(self, river_names: list[str] | str = INCLUDE_ALL_RIVER_NAMES)
             Defaults to "all".
 
         """
-        valid_river_names = list(self.indices.keys())
+        if self.indices is None:
+            valid_river_names = []
+        else:
+            valid_river_names = list(self.indices.keys())
```
ankona, line 709: extraneous `.keys()`

roms_tools/setup/river_forcing.py
```diff
@@ -806,7 +822,11 @@ def plot(
             Defaults to "all".
 
         """
-        valid_river_names = list(self.indices.keys())
+        if self.indices is None:
+            valid_river_names = []
```
ankona, line 826: consider avoiding nesting
```
valid_river_names = []
if self.indices:
    valid_river_names.extend(self.indices)
```

ankona (approved):
Looks pretty good. 

I am not a big fan of the "pass an argument named `verbose` around to decide to log". I wouldn't block this PR for it, but it really is a poor standard to set. Consider using the debug log level instead. You can even instantiate new loggers and have configuration for them if you need fine-grained control.

roms_tools/setup/nesting.py
```diff
@@ -94,14 +105,24 @@ def _map_child_boundaries_onto_parent_grid_indices(self):
...
 
         Uses a weighted sum based on boundary distance and clips depth values to a
         minimum.
         """
+        if verbose:
+            start_time = time.time()
+            logging.info("=== Modifying child topography and mask ===")
+
         # Prepare parent and child grid datasets by adjusting longitudes for dateline crossing
         parent_grid_ds, child_grid_ds = self._prepare_grid_datasets()
```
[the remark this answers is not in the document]
ankona: alternatively, use a context manager:

```python
class Timed:
    def __enter__(self):
        self.start = time.time()
        logging.info(...)

    def __exit__(self):
        logging.info("duration is: {time.time() - self.start}")
    
...
with timed(verbose):
    self._prepare_grid_datasets, verbose
```

roms_tools/setup/datasets.py
```diff
@@ -635,6 +644,7 @@ def choose_subdomain(
 
             if concats:
                 end = "both" if len(concats) == 2 else concats[0]
+                end = cast(Literal["lower", "upper", "both"], end)
```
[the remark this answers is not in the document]
NoraLoose: bd6bda43

roms_tools/setup/grid.py
```diff
@@ -796,26 +797,27 @@ def to_yaml(self, filepath: str | Path) -> None:
...
-            The name of the YAML section containing the grid configuration. Defaults to "Grid".
         verbose : bool, optional
             Indicates whether to print grid generation steps with timing. Defaults to False.
+        **kwargs : Any
 
         Returns
         -------
         Grid
             An instance of the Grid class initialized with the parameters from the YAML file.
+        **kwargs : Any
```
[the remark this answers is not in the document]
NoraLoose: b0c926d6

roms_tools/setup/grid.py
```diff
@@ -828,6 +830,8 @@ def from_yaml(
         Issues a warning if the ROMS-Tools version in the YAML header does not match the
         currently installed version.
         """
+        section_name: str = kwargs.pop("section_name", "Grid")
```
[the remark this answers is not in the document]
NoraLoose: But this will just use "Grid" if "section_name" is missing in `kwargs`, correct?
```
section_name: str = kwargs.pop("section_name", "Grid")
```
The only key ever used in `kwargs` is `section_name`. I had to introduce `kwargs` because there is a child class that inherits from `Grid` and that does not have `section_name` in the `from_yaml` method and mypy complained.

roms_tools/setup/nesting.py
NoraLoose: But there is another method `update_topography` that also calls `_map_child_boundaries_onto_parent_grid_indices` and `update_topography` has the `verbose` arguments. 

The user could do the following:
1. `ChildGrid(verbose=True, ...)`
2. `ChildGrid.update_topography(verbose=False)`

roms_tools/setup/nesting.py
NoraLoose: 9ca7046

roms_tools/setup/nesting.py
NoraLoose: Oof, this syntax is inherited from ucla. Changing it would be a breaking change for the user.

roms_tools/setup/grid.py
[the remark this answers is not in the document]
ankona: yes! i missed the default.

```python
section_sources =  {'section_name','Grid'}
if not section_sources.intersection(kwargs):
  raise RuntimeError("neither the preferred or fallback are here...")

section_name = kwargs.pop(*section_sources)
```

roms_tools/setup/nesting.py
ankona: understood

roms_tools/setup/grid.py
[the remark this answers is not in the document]
NoraLoose: 1. Can we use `*section_sources` with `pop`. `pop` takes a single key.
2. Sets are unordered, so relying on `intersection` then `pop(*set)` may be unsafe?

Maybe we can do something like this?
```
section_sources = {"section_name", "Grid"}

# Find first key that exists in kwargs
for key in section_sources:
    if key in kwargs:
        section_name = kwargs.pop(key)
        break
else:
    raise RuntimeError("Neither 'section_name' nor fallback 'Grid' is in kwargs")
```

roms_tools/setup/nesting.py
```diff
@@ -48,6 +49,8 @@ class ChildGrid(Grid):
 
         - `"prefix"` (str): Prefix for variable names in `ds_nesting`. Defaults to `"child"`.
         - `"period"` (float): Temporal resolution for boundary outputs in seconds. Defaults to 3600 (hourly).
+    verbose: bool, optional
+        Indicates whether to print grid generation steps with timing. Defaults to False.
```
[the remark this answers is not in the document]
NoraLoose: See comment in main conversation below.

roms_tools/setup/nesting.py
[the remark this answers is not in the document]
NoraLoose: that's great. a1e8415a

roms_tools/setup/river_forcing.py
```diff
@@ -474,7 +474,10 @@ def _get_overlapping_rivers(self) -> dict[tuple[int, int], list[str]]:
             A dictionary mapping grid cell indices (eta_rho, xi_rho) to a list
             of river names that overlap at that grid cell.
         """
-        index_to_rivers = defaultdict(list)
+        if self.indices is None:
+            return {}
+
+        index_to_rivers: dict[tuple[int, int], list[str]] = defaultdict(list)
```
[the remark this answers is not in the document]
NoraLoose: 0fb83a0c

roms_tools/setup/river_forcing.py
```diff
@@ -806,7 +822,11 @@ def plot(
             Defaults to "all".
 
         """
-        valid_river_names = list(self.indices.keys())
+        if self.indices is None:
+            valid_river_names = []
+        else:
+            valid_river_names = list(self.indices.keys())
```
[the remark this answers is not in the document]
NoraLoose: 90e2f6cf

roms_tools/setup/river_forcing.py
```diff
@@ -474,7 +474,10 @@ def _get_overlapping_rivers(self) -> dict[tuple[int, int], list[str]]:
             A dictionary mapping grid cell indices (eta_rho, xi_rho) to a list
             of river names that overlap at that grid cell.
         """
-        index_to_rivers = defaultdict(list)
+        if self.indices is None:
```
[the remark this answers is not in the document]
NoraLoose: The problem is that the `indices` parameter is optional, so we have
```
indices: dict[str, list[tuple[int, int]]] | None = None
```
If the user does not provide the `indices`, `__post_init__` will determine the `indices` automatically. In the other methods, I had to add the `if self.indices is None` check to make `mypy` happy.

roms_tools/setup/nesting.py
ankona: 👍

PR #407, opened 2025-08-11
Restore compatibility with `partit` 

dafyddstephenson:
This PR restores full compatibility with `Tools-Roms/partit` for `roms-tools`-generated input data.

This PR adds a subfunction `_patch_1d_encodings` to `save_datasets`. This function modifies the dataset to be saved in place such that encodings that are incompatible with `Tools-Roms/partit` (e.g. `int64`, `str`)  are replaced with ones that are (e.g. `int32`/`float64`, `char`). This purely affects on-disk encoding, only impacts some 1D variables (purely to minimise footprint, as all the impacted variables seem to be 1D), and does not change the dataset the user sees with `xr.open_dataset`. 

Variables that are true integers (like `nriver`) are saved as `int32`, variables that happen to be _encoded_ as integers (like `abs_time`) are encoded as `float64`. Variable-length string arrays are encoded as `char` arrays.

The only files that still remain incompatible with `partit` are CDR forcing files, because these should not need to be partitioned in the first place. I've opened https://github.com/CWorthy-ocean/ucla-roms/issues/50 to address that (Scott is on the case).

- Haven't added a specific test (`save_datasets` doesn't seem to be tested in isolation)
- Haven't added to `releases.md` (seems out of step with other things listed in there)

### DIFF
--- a/roms_tools/utils.py
+++ b/roms_tools/utils.py
@@ -366,7 +366,44 @@ def save_datasets(dataset_list, output_filenames, use_dask=False, verbose=True):
     List[Path]
         A list of Path objects for the filenames that were saved.
     """
+
+    def _patch_1d_encodings(dataset_list: list[xr.Dataset]) -> None:
+        """Replaces problematic encodings in 1D variables.
+
+        ROMS' Fortran-based tools fail with certain encoding types that are common
+        in roms-tools' exported 1D vars (e.g. `abs_time`, `river_name`). This function
+        replaces int64 -> int32 (for true integers), int64 -> float64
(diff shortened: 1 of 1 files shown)


dafyddstephenson:
May want to extend the protection here to cover >1D vars as well. I don't think the encoding patch does a full read, so it should be fine performance wise, but having it as just 1D vars up to now made double sure.

NoraLoose:
> I was hoping to see how much slowdown we get by just applying the patch to all vars first, might be a better safety net

We could also open a separate issue to investigate this. I was hoping to merge this PR soon, because I need to fix some other encoding-related stuff: https://github.com/CWorthy-ocean/roms-tools/issues/415

PR #413, opened 2025-08-14
Centered gaussian release

smaticka:
This PR reflects the changes being done in roms related to this isssue: https://github.com/CWorthy-ocean/ucla-roms/issues/31

Previously, the gaussian shape was previously multiplied by the height of each cell, but since the cell heights varied, the larger cells reflected more mass.

By removing the cell height product, the distribution looks more centered around the release location:

image

### DIFF
--- a/.github/workflows/sphinx.yaml
+++ b/.github/workflows/sphinx.yaml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Create conda environment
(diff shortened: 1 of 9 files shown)


smaticka:
Thanks for adding the details to `release.md` @NoraLoose. I forgot about that pre-commit step. Do you think it's appropriate to add the things to do when contributing? here: https://github.com/CWorthy-ocean/roms-tools?tab=contributing-ov-file

I think this PR is otherwise ready to go.

Issue #415, opened 2025-08-18
Unpin xarray version


NoraLoose:
In some of the open PRs, CI fails with the following error:

```
FAILED roms_tools/tests/test_setup/test_boundary_forcing.py::test_boundary_forcing_save[boundary_forcing] - ValueError: Unlimited dimension(s) {'time'} declared in 'dataset.encoding', but not part of current dataset dimensions. Consider removing {'time'} from 'dataset.encoding'.
```

dafyddstephenson:
Is everything now patched?

PR #416, opened 2025-08-19
ROMS-415: Pin xarray version

ankona:
This PR includes a temporary fix for test failures when installing `xarray==2025.8.0`

Workaround for issue #415 


NoraLoose:
Can we merge this? Or are the dask tests hanging because of the pinned xarray version? (I don't think so.)

NoraLoose:
We need #425 for the tests to pass.

PR #436, opened 2025-08-26
Skip streaming tests on forks

NoraLoose:
... because we don't have the `copernicusmarine` secrets on forks, which is why the streaming tests fail there.


ankona (approved):
Thanks for cleaning up the mess I left in your house!

ankona (changes requested):
I think you want to do `if: github.repository == github.event.pull_request.head.repo.full_name` instead.

The repository_owner should be `C-Worthy` and this won't work in the main repo.

PR #460, opened 2025-09-10
Safe mask selection for GLORYS dataset

NoraLoose:
- [ ] Closes #xxxx
- [ ] Tests added

### DIFF
--- a/roms_tools/setup/datasets.py
+++ b/roms_tools/setup/datasets.py
@@ -1073,19 +1073,29 @@ class GLORYSDataset(Dataset):
         None
             The dataset is modified in-place by applying the mask to each variable.
         """
-        mask = xr.where(
-            self.ds[self.var_names["zeta"]].isel({self.dim_names["time"]: 0}).isnull(),
-            0,
-            1,
-        )
-        mask_vel = xr.where(
-            self.ds[self.var_names["u"]]
-            .isel({self.dim_names["time"]: 0, self.dim_names["depth"]: 0})
-            .isnull(),
-            0,
-            1,
-        )
+        zeta = self.ds[self.var_names["zeta"]]
+        u = self.ds[self.var_names["u"]]
 
+        # Select time=0 if time dimension exists, otherwise use data as-is
+        if self.dim_names["time"] in zeta.dims:
+            zeta_ref = zeta.isel({self.dim_names["time"]: 0})
+        else:
+            zeta_ref = zeta
+
+        if self.dim_names["time"] in u.dims:
+            u_ref = u.isel({self.dim_names["time"]: 0})
+        else:
+            u_ref = u
+
+        # Also handle depth for velocity
+        if self.dim_names["depth"] in u_ref.dims:
+            u_ref = u_ref.isel({self.dim_names["depth"]: 0})
+
+        # Create masks
+        mask = xr.where(zeta_ref.isnull(), 0, 1)
+        mask_vel = xr.where(u_ref.isnull(), 0, 1)
+
+        # Save to dataset
         self.ds["mask"] = mask
         self.ds["mask_vel"] = mask_vel
 


roms_tools/setup/datasets.py
```diff
@@ -2764,138 +2795,209 @@ def _check_dataset(
...
-                logging.info(
-                    f"Selected time entry closest to the specified start_time ({start_time}) within the range [{start_time}, {start_time + timedelta(hours=24)}]: {ds[time_dim].values}"
-                )
     else:
-        logging.warning(
-            "Dataset does not contain any time information. Please check if the time dimension "
-            "is correctly named or if the dataset includes time data."
+        # Assume we are looking for exactly one time record for initial conditions
+        ds = _select_initial_time(
+            ds, time_dim, start_time, climatology, allow_flex_time
```
ScottEilerman, line 2930: nit / style / optional: you could move this statement up and out with
```python
if not end_time:
    return _select_initial_time(ds, time_dim, start_time, climatology, allow_flex_time)

# ... rest of logic currently in the if statement
```

Then you don't need an else and you get to "unindent" the bulk of the function which can sometimes help with readability.

roms_tools/setup/datasets.py
```diff
@@ -2764,138 +2795,209 @@ def _check_dataset(
...
+
+    Raises
+    ------
+    ValueError
+        If no matching time is found (when `allow_flex_time=False`), or no entries are
+        available within the +24h window (when `allow_flex_time=True`).
+    """
+    if climatology:
+        # Convert from timedelta64[ns] to fractional days
+        ds["time"] = ds["time"] / np.timedelta64(1, "D")
```
ScottEilerman, line 2969: I assume since this was already in the code, just moved, that it works fine. I'd guess the origin date would be important here, but probably that's already accounted for somewhere? Figure I'd ask just in case.

roms_tools/setup/datasets.py
```diff
@@ -2764,138 +2795,209 @@ def _check_dataset(
...
+        # Strict match required
+        if not (ds[time_dim].values == np.datetime64(ini_time)).any():
+            raise ValueError(
+                f"No exact match found for initial time {ini_time}. Consider setting allow_flex_time to True."
+            )
+
+        ds = ds.sel({time_dim: np.datetime64(ini_time)})
+
+    if time_dim not in ds.dims:
+        ds = ds.expand_dims(time_dim)
```
ScottEilerman, line 3000: Does this condition only happen because of the `sel`s above, and we want to add it back? Or is it possible we've been fed a ds without `time_dim` in it? I assume that would break the above blocks.

ScottEilerman (approved):
Really minor comments and questions. Fine as-is if none of the questions trigger any concerns for you!

roms_tools/setup/datasets.py
```diff
@@ -2764,138 +2795,209 @@ def _check_dataset(
...
-                # Convert from timedelta64[ns] to fractional days
-                ds["time"] = ds["time"] / np.timedelta64(1, "D")
-                # Interpolate from climatology for initial conditions
-                ds = interpolate_from_climatology(ds, time_dim, start_time)
+            return ds
+
+        # Identify records before or at start_time
+        before_start = ds[time_dim] <= np.datetime64(start_time)
+        if before_start.any():
+            closest_before_start = ds[time_dim].where(before_start, drop=True).max()
```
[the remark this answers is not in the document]
NoraLoose: [73e8a8b](https://github.com/CWorthy-ocean/roms-tools/pull/460/commits/73e8a8b79e71681678ec1f764b772fa987712942)

roms_tools/setup/datasets.py
NoraLoose: 65e5fe1f

roms_tools/setup/datasets.py
NoraLoose: In the next step, we call `interpolate_from_climatology`, which takes the 12 climatological values and interpolates the `ini_time` from them. For this operation, the original reference date (`model_reference_date`) is not needed.  

Later, within `InitialConditions` (but not in the generic `Dataset` class, which only handles the raw datasets without ROMS-specific logic), the `ini_time` is then converted to relative days since `model_reference_date`.

Maybe I didn't get your question right?

roms_tools/setup/datasets.py
NoraLoose: It only happens because of the `sel`s above. That is what broke the code somehow. 

https://github.com/CWorthy-ocean/roms-tools/blob/2767c497f9fa58ac324b86a0caa56c66eab787a5/roms_tools/setup/datasets.py#L140-L142 makes sure that we never feed it a `ds` without `time_dim`.

PR #487, opened 2025-11-19
Remove coordinates from nesting info file

### LINKED ISSUE #464: Error when nesting using the nesting_info.nc file
I get the following error, when I try to do nesting with roms-tools:

`       extract_data :: read objects: Iceland1_edata.nc

   ERROR::
       40
            ncread read var= indices_rho

   NetCDF: Attempt to convert between text & numbers`

NoraLoose:
I removed the coordinates `indices_rho` and `indices_vel` from the nesting info (or "edata") file because ROMS doesn't like these additional metadata.

- [x] Closes #464

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -29,6 +29,7 @@
 
 * Fix bug in creation of `InitialConditions` ([#460](https://github.com/CWorthy-ocean/roms-tools/pull/460))
 * Remove coordinate variables in nesting info file that conflict with ROMS expectations ([#487](https://github.com/CWorthy-ocean/roms-tools/pull/487))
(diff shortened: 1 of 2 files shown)


NoraLoose:
Thanks @ubbu36!

> However, when applying extract_data_join on the *.ext the files, the name of the resultant boundary file is child_bry.20241006000240.nc. I'm wondering whether we can make it more explicit how to determine the name of the boundary file so that the user can keep track of the different nests and their boundary files. In the MATLAB script bry_extract_obj.m, the bry files are named from the variable 'gname' I believe

To address this, I will include some explanation in the documentation in a follow-up PR.

PR #491, opened 2025-11-20
Allow child domains with land points outside parent

### LINKED ISSUE #480: Allow child domains with land points outside parent
It might be desirable to have child grids with boundaries outside the parent grid, if those points are all land.

Currently, this situation throws an error here:
https://github.com/CWorthy-ocean/roms-tools/blob/2830da6461d78ba2f3a097b3666dd4215bb39745/roms_tools/setup/nesting.py#L503

It seems like an easy fix could be to add something like:
```
lon_masked = lon.where(mask)
lat_masked = lat.where(mask)
```
before making the call to `griddata`.

NoraLoose:
## PR Description

This PR allows child domains to extend beyond the parent domain **as long as those out-of-bounds points are land in the child grid**.

Screenshot 2025-11-24 at 10 33 50 AM

### Changes

- **Reordered operations in `ChildGrid` for correctness and clarity**  
  New sequence:
  
  1. Create horizontal grid  
  2. Derive mask from coastlines  
  3. Map child grid boundary points to parent grid indices  
  4. Modify the child mask  
  5. Generate topography  
  6. Modify the child topography  
  7. Compute vertical coordinate system

  **Notes:**
  - Steps 2, 3, 4 belong to `update_mask`.  
  - Steps 5 and 6 belong to `update_topography`.  
  - Mapping boundaries (step 3) depends on the updated mask (step 2) because child land points may extend beyond the parent grid.  
  - The modification steps (4 and 6) apply only to valid parent–child configurations and must occur *after* boundary mapping.

- **Fixed redundant boundary mapping and cleaned up initialization**  
  Previously, boundary mapping was run twice, and `ChildGrid` unnecessarily overwrote `__post_init__`. Both issues are resolved.

- **Refactored mask/topography modification**  
  Split `_modify_child_topography_and_mask` into:
  - `_modify_child_mask`
  - `_modify_child_topography`

- **Updated `ChildGrid.plot_nesting`**  
  Now supports nesting configurations where the child grid extends beyond the parent domain.

- **Added `apply_mask` option to `plot`**  
  Enables plotting mask differences using the main plotting interface (with proper map projections), instead of relying on xarray’s default plotting.

- **Updated documentation and tests**

### Checklist

- [x] Closes #480
- [x] Tests added

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -60,8 +60,8 @@ jobs:
           name: without-dask-py${{ matrix.python-version }}
           fail_ci_if_error: false
 
-  test-with-dask:
-    name: 3.13-build-with-dask
+  test-with-dask-1:
+    name: 3.13-build-with-dask-1
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -93,7 +93,7 @@ jobs:
         shell: bash -l {0}
         run: |
            python -V
-           coverage run --rcfile=coverage.toml -m pytest --use_dask --verbose roms_tools/tests/*
+           coverage run --rcfile=coverage.toml -m pytest --use_dask --verbose roms_tools/tests/ --ignore=roms_tools/tests/test_setup/test_boundary_forcing.py --ignore=roms_tools/tests/test_setup/test_initial_conditions.py --ignore=roms_tools/tests/test_setup/test_surface_forcing.py
 
       - name: Get coverage report
         shell: bash -l {0}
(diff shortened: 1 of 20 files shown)


roms_tools/setup/nesting.py
```diff
@@ -492,23 +560,31 @@ def interpolate_indices(parent_grid_ds, lon, lat, mask):
...
+        or j.where(mask).min() < 0
+        or j.where(mask).max() > nyp - 2
+    ):
         logging.warning(
-            "Some boundary points of the child grid are very close to the boundary of the parent grid."
+            "Some ocean child boundary points lie very close to the edges of the parent grid."
         )
 
+    # Fill NaNs (land child points that fall outside parent grid) with fill value
+    i = i.fillna(-1e5)
```
ScottEilerman: What's the meaning of -1e5 here? I haven't really looked at what these arrays get used for, to know how -1e5 will get interpreted later. I guess my assumption just from reading the docstring in this function is that this effectively places these off-grid land points "really far away" such that they won't get used for interpolation later?

NoraLoose: Good question. The choice of `-1e5` comes from the original [UCLA MATLAB scripts](https://github.com/nmolem/ucla-tools/blob/ff9a181b70ecf8918f46555a9508c7f30c5bf56b/edata/add_object.m#L57-L58).

I was also initially skeptical about using `-1e5` as a fill value. The reason we need it now is that `update_indices_if_on_parent_land` will fail if there are NaNs present. Previously, this wasn’t an issue because child grids were always fully contained within the parent grid. (In fact, I was initially concerned that filling and interpolating could create edge-case issues, which is why this scenario was excluded before this PR.)

Looking at the MATLAB scripts again, it seems we are not fully reproducing the section that attempts to fix NaNs before resorting to the `-1e5` fill value: https://github.com/nmolem/ucla-tools/blob/ff9a181b70ecf8918f46555a9508c7f30c5bf56b/edata/add_object.m#L31-L38.

I'll make another push to address this.

To be honest, I’m not 100% confident I understand all of the MATLAB implementation yet.

NoraLoose: I'm not convinced that `1e-5` would be read in as a NaN, as Jeroen was concerned about. I could go either way. What error message would you throw?

roms_tools/setup/nesting.py
```diff
@@ -569,8 +569,30 @@ def interpolate_indices(
...
-    # Warn if child boundary points are near the edges of the parent grid
     nxp, nyp = lon_parent.shape
+    idx = xr.DataArray(np.arange(i.size), dims=i.dims)
+    nan_idx = (
+        i.isnull() | j.isnull() | (i > nxp - 2) | (i < 0) | (j > nyp - 2) | (j < 0)
+    )  # remaining NaNs and out-of-bound indices
+
+    # Interpolate indices for points that are invalid (NaN or out-of-bounds),
+    # but only if there are some valid points to use for interpolation.
+    if nan_idx.any() and not nan_idx.all():
```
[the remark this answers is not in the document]
NoraLoose: The `.all()` case is when the user has requested a boundary that lies entirely on land. Do we want to tell the user to exclude that boundary? Maybe.

roms_tools/setup/nesting.py
NoraLoose: Maybe in most cases this case means that the user has requested a boundary that lies entirely on land. We could suggest that the user excludes that boundary. Maybe there are other edge cases, too, though.

NoraLoose:
@ScottEilerman We are now avoiding the fill value altogether. What do you think?

PR #504, opened 2025-12-08
Default boundary selection respects land-only boundaries

### LINKED ISSUE #502: Default boundary selection should respect land-only boundaries
Both `BoundaryForcing` and `ChildGrid` take a `boundaries` dictionary, for example:

```python
boundaries={
    "south": True,
    "east": True,
    "north": False,
    "west": True,
}
```
If the user does not provide this argument, the current default is:
```
{"south": True, "east": True, "north": True, "west": True}
```
### Suggested change

Automatically set any boundary to `False` when it is fully land-only.

### Backwards compatibility

If the user explicitly provides boundaries, their settings always take precedence.

NoraLoose:
Closes #502. I had to overwrite some of the test data because a smaller set of boundary data was produced for some of the test domains.

- [x] Closes #502
- [x] Tests added

### DIFF
--- a/roms_tools/setup/utils.py
+++ b/roms_tools/setup/utils.py
@@ -2321,6 +2321,6 @@ def _infer_valid_boundaries_from_mask(mask: xr.DataArray) -> dict[str, bool]:
         bdry_mask = mask.isel(**coords)
 
         # Boundary is valid if ANY ocean point exists
-        boundaries[direction] = bool(bdry_mask.sum() > 0)
+        boundaries[direction] = bool(bdry_mask.values.any())
 
     return boundaries


roms_tools/setup/utils.py
```diff
@@ -2224,3 +2224,103 @@ def _select_initial_time(
...
+    """
+    bdry_coords = get_boundary_coords()
+    boundaries = {}
+
+    for direction in ["south", "east", "north", "west"]:
+        coords = bdry_coords["rho"][direction]
+        bdry_mask = mask.isel(**coords)
+
+        # Boundary is valid if ANY ocean point exists
+        boundaries[direction] = bool(bdry_mask.sum() > 0)
```
ScottEilerman: Minor optimization note: `bdry_mask.values.any()` might be able to replace `bool(bdry_mask.sum() > 0)` here and avoid summing an entire array if, say, the first value is ocean.

NoraLoose: Cool. I implemented your suggestion in [d0793e2](https://github.com/CWorthy-ocean/roms-tools/pull/504/commits/d0793e2231c852728478d6823eb5039bfe884bf8).

PR #506, opened 2025-12-09
Improved bathymetry smoothing to better match UCLA MATLAB tools

NoraLoose:
This PR updates the bathymetry smoothing in two ways, aiming to bring the results closer to those produced by the UCLA MATLAB tools:

* **Domain-wide smoothing**: The smoothing factor has been reduced from `6.0` to `5.0`.

* **Local smoothing**: Previously, masking was applied before smoothing. Now, smoothing is applied directly without masking, which noticeably improves results near islands.

@smaticka Could you add a few figures to illustrate how these changes bring the results closer to the UCLA outputs?


smaticka:
Thanks @NoraLoose. These changes look good to me. 

This is the bathymetry difference between the domain of UCLA and our domain _before the changes_ (max bathy difference of 3753 m):
image

And these are the differences _after the changes from this PR_ (max difference is 318 m):
image

These are for the Atlantic domain that looks like this:
image

PR #550, opened 2026-02-05
Improve `ChildGrid` boundary defaults and add `include_bgc` metadata option

### LINKED ISSUE #548: No BGC in nesting?
I can't find a way to prepare the nesting such that the file produced by `child_grid.save_nesting` contains metadata that instructs ROMS to include BGC in the extracted boundary forcing. Looking at other `edata` files, it seems as though wherever we have 

```
output_vars = "zeta, temp, salt" ;
```

in the metadata, adding `bgc` as follows is sufficient:

```
output_vars = "zeta, temp, salt, bgc" ;
```

Hopefully I'm either missing something or it's a quick fix! :)

NoraLoose:
### Summary

This PR improves the nesting functionality in the `ChildGrid` class in two ways:

1. **Automatic boundary detection**  
   If no `boundaries` argument is provided, open boundaries are now inferred from the grid mask rather than defaulting all boundaries to `True`. This behavior was the original intention of #504 but was not fully implemented; it is now fixed and covered by tests.

2. **Configurable BGC output via metadata**  
   Users can now specify `"include_bgc"` in the `metadata` dictionary to control whether biogeochemical variables are included in boundary outputs.

3. **Validation test for `ChildGrid`**  
   The fixture `child_grid_with_bgc` has been added to the regression check suite to ensure that future changes do not unintentionally alter results.

### Notes

If I were to design `ChildGrid` again today, I would represent `metadata` as a dedicated dataclass instead of a free-form dictionary, which would be more robust and type-safe. However, this PR keeps the existing dictionary-based interface to preserve backward compatibility.

- [x] Closes #548
- [x] Tests added


ScottEilerman (approved):
One comment about some maybe-leftover text in the docs, otherwise LGTM!

PR #559, opened 2026-03-09
ensure ChildGrid mask is int32

### LINKED ISSUE #558: `ChildGrid.save` produces netCDF file incompatible with `Tools-Roms/partit`
Seems that `update_child_mask` recasts `mask_rho` as `int64`, which is not supported by the Fortran `partit` tool

dafyddstephenson:
netCDF created by `ChildGrid.save` is otherwise incompatible with `Tools-Roms/partit`. Now consistent with parent `Grid.save`

Closes #558


NoraLoose:
Could you end a one-liner to the release notes? Thanks! :)

PR #573, opened 2026-03-25
Child grid refactoring

### LINKED ISSUE #490: `Grid.to_yaml()` produces mixed inferred/Nones for externally created grids
### Summary

For externally created grids (e.g., MATLAB-generated grids), `Grid.from_file()` successfully produces a `Grid` object. However, calling `.to_yaml()` on such grids outputs a YAML that mixes inferred parameters and `None` values:

```yaml
---
roms_tools_version: 3.1.2.dev23+gc2b82bc5c
---
Grid:
  nx: 24
  ny: 24
  size_x: null
  size_y: null
  center_lon: -4.1
  center_lat: 52.4
  rot: 0.0
  N: 100
  theta_s: 5.0
  theta_b: 2.0
  hc: 300.0
  topography_source: null
  hmin: null
```
Inferring all parameters in this case is dangerous because the user could then use the YAML to create a ROMS-Tools grid that differs from the original MATLAB grid.

### Proposed behavior

* If all key parameters are known, output the full YAML (current preferred behavior).
* If any required parameter is missing, output only the path to the original grid file, e.g.:
```
---
roms_tools_version: 3.1.2.dev23+gc2b82bc5c
---
Grid:
  path: path_to_grid
```
This avoids misleading users and preserves the integrity of externally created grids.

smaticka:
A number of changes were done to address the refactoring. See issue #571 for a detailed breakdown.

The ChildGrid class was removed and the 2 main functions replacing its methods are `align_grid` and `make_edata`. 

Since the order of operations was changed from the current 'generate edata, then update mask' to 'update mask, then generate edata', an extra check was needed to confirm the child grid wet points aren't outside the parent domain (issue #491). The check added in this might not be entirely robust. It did not handle well grids that cross the prim meridian and change from 350s longitude to 0-10s longitude. 

The `Grid` class now allows for loading of a netcdf grid file by calling it's private function `_from_file()` e.g. `Grid('grid_12km.nc')`. No updates are made to a grid that is loaded via this method in order to preserve any grids that were modified or generated external to ROMS-Tools.  If a Grid is loaded from an external file, the `to_yaml` method of Grid will now store just the filepath instead of the usual parameters (see issue #490). 

The new method for making a child gird is to use the Grid class, and then run `align_grid` to smooth the topography at the boundaries to match that of the parent grid. At this point, if the child grid uses the Grid class `to_yaml` `from_yaml` method, the Grid returned `from_yaml` will return the child grid in the state _before_ it was altered by `align_grid`. A decision needs to be made to address what the behavior should be. 

There is no function to read in an edata file that is created using make_edata(). The user can load it in with xarray if desired. 

The long_name and units were missing from various variables in the ChildGrid construct, this PR addressed some of them but does not addresses the missing coordinates for `lat_rho, lat_u, lat_v`. They appear to be lost when `wrap_longitude` is called under `_apply_child_modifications` > `_prepare_grid_datasets` 

Resolves #571 
Resolves #491 
Resolves #490 

- [x] Update other notebooks that call Grid.from_file()
- [ ] make issue for missing coordinates
- [x] Need to update nesting notebook for include_pressure_fluxes
- [x] Need to update the to_yaml from_yaml to retain childgrid boundaries
- [x] Need to fix the checking of wet child points outside of the parent domain. 
- [x] Tests adjusted


ScottEilerman (approved):
Reviewed and iterated via pull requests and sitting in the same room. Approved!

PR #577, opened 2026-03-27
Fix enclosed basin labeling and update connectivity test

ubbu36:
**Summary**
Fixes enclosed-basin masking so all labeled wet regions are handled correctly.
**Changes**
Fixed _fill_enclosed_basins loop bounds to iterate component labels 1..nreg (instead of skipping label nreg).
Updated test_enclosed_regions to assert a single connected wet region (nreg == 1) after basin filling.
**Why**
The previous off-by-one logic could leave a small enclosed wet region unprocessed, creating spurious interior lakes.


NoraLoose (approved):
Looks great, thanks for fixing this bug!

For the paper trail, here is the connectivity before and after this fix:

Before:
Image

After:
Image

Note the small yellow island in the before plot that is now removed.

PR #582, opened 2026-04-10
Default chunking, initial slicing 

blsaenz:
The last PR on dask dataset reading relies on an underlying netcdf/HDF file's chunking to perform optimized reads and processing.  This one sets default lateral chunking that dask forces upon large arrays, which allows dask to read just what is needs from un-chunked underlying files (like Glorys and roms output).  Things go faster now and use less memory using existing files.   This behavior can be overridden in existing classes by passing in custom chunking.  

There is also a related option to pass in 'initial_slice_bounds' in the form of lat/lon bounds to classes that use lat_lon_dataset.  This information is passed to xarray which does initial slicing BEFORE handing the data to dask.  In testing, this method for BC and initial conditions resulted in fastest file creation and lowest memory usage (compared to above fix).


blsaenz:
Huh, this branch was copied from another and has all the old history that has been merged with the last PR.  Sorry

blsaenz:
I added a couple of pytests, maybe not enough? Trying to think of a way to test dask chunking differences but it doesn't make a ton of sense.

blsaenz:
Tests not happy. I guess we need to talk about what should be the default behavior for dask - 1) chunks or 2) no chunks.  1) we change the tests to load data overriding the default chunking that cause failed tests, because they are not expecting dask chunks, 2) the user is going to be make to pass in chunks OR we have default chunking on the outer BoundaryCondition classes etc.

Thinking about this more ... it seems reasonable if we have a use_dask switch, that a pytests/users should not expect the dataset to be un-chunked when returned.  E.g. a use_dask test should be able to handle dask chunks. So, I'm leaning toward 1) above.  

Edit: It wasn't so bad.  Most fails were caused by the shortwave irradiance correction using a custom method for subsetting (which didn't unchunk like in choose_subdomain()). A handful tests for dataset initialization needed unchunking in the pytests.

roms_tools/datasets/lat_lon_datasets.py
```diff
@@ -52,6 +52,8 @@
 DEFAULT_NR_BUFFER_POINTS = (
     20  # Default number of buffer points for subdomain selection.
 )
+# Default lateral chunk size for Dask-backed LatLonDataset subclasses (latitude/longitude).
+_DEFAULT_LAT_LON_LATERAL_CHUNK = 50
```
NoraLoose: Better to move this line down to after 61 to not break up the comment for the previous global variable

blsaenz: Done

roms_tools/utils.py
```diff
@@ -217,6 +225,120 @@ def get_dask_chunks(
...
+                dim: slice(pair[0], pair[1])  # type: ignore[arg-type]
+                for dim, pair in bounds.items()  # type: ignore[arg-type]
+            }
+
+        prep_func = partial(
+            self._apply_slices,
+            selectors=self.selectors,
+            use_isel=self._use_isel_for_initial_slice,
+        )
+
```
ScottEilerman, line 261: Refactoring consideration (feel free to ignore if you want to move on): I'm not sure I understand the need for the class here. I think you could make `_apply_slices` a regular function (it already effectively is, since it's static), then make lines 294-313 here a function called `_get_ds_preprocessor()` with only the args needed for this portion. Return the preprocessor (`prep_func`) from that function.

```python

def _apply_slices(ds, selectors, use_isel: bool = False):
    """
    The method passed to 'preprocess'.
    It checks which dimensions exist in the current file before slicing.
    """
    # Intersection: only use keys that exist in the current file's dimensions
    # This prevents 'KeyError' if some files have different dim names
    active_selectors = {k: v for k, v in selectors.items() if k in ds.dims}

    if active_selectors:
        if use_isel:
            return ds.isel(**active_selectors)
        return ds.sel(**active_selectors)

    return ds

def _get_ds_preprocessor(initial_slice_bounds, initial_slice_bounds_use_isel):
    bounds = initial_slice_bounds or {}
    if initial_slice_bounds_use_isel:
        selectors = {
            dim: slice(int(pair[0]), int(pair[1]) + 1) for dim, pair in bounds.items()
        }
    else:
        selectors = {
            # type checkers failing here (wants only int), but slice can actually take any type
            dim: slice(pair[0], pair[1])  # type: ignore[arg-type]
            for dim, pair in bounds.items()  # type: ignore[arg-type]
        }

    prep_func = partial(
        _apply_slices,
        selectors=selectors,
        use_isel=initial_slice_bounds_use_isel,
    )

    return prep_func
```

Then, where it gets used below (line 426), instead of having this separate class call,  conditionally get the preprocessing function, and the if/else collapses to a single `open_mfdataset` call. That also eliminates the need to "pass through" all of these args to `open_mfdataset`.

```python
        preprocessor = (
            _get_ds_preprocessor(initial_slice_bounds, initial_slice_bounds_use_isel)
            if initial_slice_bounds is not None
            else None
        )
        return xr.open_mfdataset(
            filenames,
            decode_times=decode_times,
            decode_timedelta=decode_timedelta,
            chunks=chunks,
            preprocess=preprocessor,
            **kwargs,
        )
```

ScottEilerman (commented):
One suggested structural change, but otherwise you're deeper in this than I, so if it's all working according to expectations, I'm happy to approve!

roms_tools/setup/boundary_forcing.py
```diff
@@ -96,6 +96,28 @@ class BoundaryForcing:
         Indicates whether to skip validation checks in the processed data. When set to True,
         the validation process that ensures no NaN values exist at wet points
         in the processed dataset is bypassed. Defaults to False.
+    initial_slice_bounds : dict, optional
+        Optional horizontal subset to apply when loading. Two forms are supported:
+
+        - Index bounds on rho dimensions: keys are the dataset dimension names for rho
```
NoraLoose: Is this bit relevant for the BoundaryForcing, where source data comes always in lat-lon, and never in ROMS native dims?

roms_tools/setup/surface_forcing.py
```diff
@@ -101,6 +101,16 @@ class SurfaceForcing:
         Indicates whether to skip validation checks in the processed data. When set to True,
         the validation process that ensures no NaN values exist at wet points
         in the processed dataset is bypassed. Defaults to False.
+    initial_slice_bounds : dict, optional
+        Optional horizontal subset to apply when loading data via dask. One form is supported
+        for initial conditions, geographic bounds:
+
```
NoraLoose, line 112: Did you mean to write "one form is supported for surface forcing"?

roms_tools/setup/boundary_forcing.py
NoraLoose: Nit: personally, I also find the terminology "ROMS native curvilinear dimensions" more intuitive than "rho dimensions", because 
1) sometimes ROMS output doesn't even have "rho" in them if the output is 2D
2) rho indicates the vertical (terrain following) coordinate, but here the lateral (curvilinear) dims are maybe more relevant for lateral rechunking?

roms_tools/utils.py
```diff
@@ -185,26 +186,33 @@ def get_dask_chunks(
         If True and `use_dask=True`, the data will be chunked along the time dimension with a chunk size of 1.
         If False, the data will not be chunked explicitly along the time dimension, but will follow the default auto chunking scheme. This option is useful for ROMS restart files.
         Defaults to True.
+    lateral_chunk : int, optional
+        The chunk size for the lateral dimensions.
+        Defaults to -1.
+        If -1, the chunk size will be determined by the data.
```
NoraLoose: If -1, we force a single chunk per horizontal dim, correct?

roms_tools/utils.py
```diff
@@ -217,6 +225,120 @@ def get_dask_chunks(
     return chunks
 
 
+def unchunk_dask(
```
NoraLoose, line 227: What's the difference between unchunk_dask and reset_chunking? Possible to use only one of them?

roms_tools/setup/boundary_forcing.py
```diff
@@ -96,6 +96,28 @@ class BoundaryForcing:
...
+          points (usually ``"eta_rho"`` and ``"xi_rho"``, i.e. ``dim_names`` values),
+          values are inclusive ``(min_index, max_index)`` tuples.
+
+        - Geographic bounds: ``{"latitude": (min_lat, max_lat), "longitude": (min_lon, max_lon)}``
+          in degrees. Indices are computed with
```
NoraLoose: Could think about cutting it off after "degrees" to not overwhelm the user with internal details in the docstrings for this user- facing class. But these details are great to keep for the in-code comments? Maybe this is just my personal preference, so feel free to ignore

roms_tools/setup/initial_conditions.py
```diff
@@ -101,6 +101,14 @@ class InitialConditions:
         Indicates whether to skip validation checks in the processed data. When set to True,
         the validation process that ensures no NaN values exist at wet points
         in the processed dataset is bypassed. Defaults to False.
+    initial_slice_bounds : dict, optional
+        Optional horizontal subset when loading lat/lon forcing datasets (e.g. GLORYS)
+        with Dask. Not used for ROMS restart sources.
```
NoraLoose: Could be helpful to give the user an example for the dict here (similarly as done in BoundaryForcing docstrings)

roms_tools/setup/boundary_forcing.py
```diff
@@ -96,6 +96,28 @@ class BoundaryForcing:
...
+          ``self.grid`` and then passed as rho index bounds above.
+
+        Subsetting requires rho dimensions to use index-like coordinates so that
+        :meth:`xarray.Dataset.sel` can select along ``eta_rho`` / ``xi_rho`` (typical ROMS
+        NetCDF output). When ``initial_slice_bounds`` is set, the loaded dataset may be
+        smaller than ``grid.ds`` along rho horizontal dimensions; the grid object is
+        unchanged.
+    chunks : dict[str, int], optional
+        Optional Dask chunk sizes passed to
+        :class:`~roms_tools.datasets.lat_lon_datasets.LatLonDataset` subclasses when
```
NoraLoose: Similar consideration here. Too much info for user-facing docstrings? Not sure. Could put info as comment in code?

NoraLoose: In any case, we want to keep the last sentence "If None, ..."

roms_tools/setup/boundary_forcing.py
```diff
@@ -96,6 +96,28 @@ class BoundaryForcing:
...
+        loading with ``use_dask=True``. If ``None``, each dataset class applies its own
+        defaults.
```
NoraLoose: Are the defaults maybe always "a single chunk in horizontal dims"? If so, could state this here.

roms_tools/setup/initial_conditions.py
```diff
@@ -101,6 +101,14 @@ class InitialConditions:
...
+    chunks : dict[str, int], optional
+        Optional Dask chunk sizes when loading with ``use_dask=True``: passed to
+        :class:`~roms_tools.datasets.lat_lon_datasets.LatLonDataset` subclasses for lat/lon sources
+        and to :class:`~roms_tools.datasets.roms_dataset.ROMSDataset` for ROMS restart sources.
+        If ``None``, each dataset class applies its own defaults.
```
NoraLoose: Are the defaults maybe always "a single chunk in horizontal dims"? If so, could state this here.

roms_tools/setup/boundary_forcing.py
```diff
@@ -96,6 +96,28 @@ class BoundaryForcing:
...
+
+        - Geographic bounds: ``{"latitude": (min_lat, max_lat), "longitude": (min_lon, max_lon)}``
+          in degrees. Indices are computed with
+          :meth:`~roms_tools.setup.grid.Grid.rho_index_bounds_from_latlon_bounds` on
+          ``self.grid`` and then passed as rho index bounds above.
+
+        Subsetting requires rho dimensions to use index-like coordinates so that
+        :meth:`xarray.Dataset.sel` can select along ``eta_rho`` / ``xi_rho`` (typical ROMS
+        NetCDF output). When ``initial_slice_bounds`` is set, the loaded dataset may be
+        smaller than ``grid.ds`` along rho horizontal dimensions; the grid object is
```
NoraLoose: I know what you mean bere, but "rho horizontal dims" makes maybe not the most sense because rho indicates the native vertical dim

roms_tools/setup/boundary_forcing.py
```diff
@@ -96,6 +96,28 @@ class BoundaryForcing:
...
+    initial_slice_bounds : dict, optional
+        Optional horizontal subset to apply when loading. Two forms are supported:
+
+        - Index bounds on rho dimensions: keys are the dataset dimension names for rho
+          points (usually ``"eta_rho"`` and ``"xi_rho"``, i.e. ``dim_names`` values),
+          values are inclusive ``(min_index, max_index)`` tuples.
+
+        - Geographic bounds: ``{"latitude": (min_lat, max_lat), "longitude": (min_lon, max_lon)}``
+          in degrees. Indices are computed with
+          :meth:`~roms_tools.setup.grid.Grid.rho_index_bounds_from_latlon_bounds` on
```
NoraLoose: I don't see this new rho_index_bounds_from_latlon_bounds method on Grid in this PR?

roms_tools/setup/surface_forcing.py
```diff
@@ -101,6 +101,16 @@ class SurfaceForcing:
...
+        - Geographic bounds: ``{"latitude": (min_lat, max_lat), "longitude": (min_lon, max_lon)}``
+          in degrees.
+    chunks : dict[str, int], optional
+        Dask chunk sizes passed to
+        :class:`~roms_tools.datasets.lat_lon_datasets.LatLonDataset` subclasses when
+        loading with ``use_dask=True``. Pass ``None`` to default to no chunking along non-time dimensions.
```
NoraLoose: Maybe more intuitive to say "no chunking along spatial dimensions"?

roms_tools/setup/initial_conditions.py
NoraLoose: That's at least what is states in the SurfaceForcing docstrings!

roms_tools/setup/boundary_forcing.py
NoraLoose: That's at least what is states in the SurfaceForcing docstrings!

NoraLoose:
Nice work, @blsaenz! It would be very helpful to add some examples to the documentation for how to use this new feature!

NoraLoose:
One minor comment: You could consider moving the two new input parameters directly under use_dask and before bypass_validation, so that the dask-related parameters appear all together in the API reference.

NoraLoose:
> It would be very helpful to add some examples to the documentation for how to use this new feature!

I think this could fit nicely in the "Using Dask" notebook in the docs. Maybe you could add a new section there that explains what the new parameters do. Could also do some timing and memory profiling comparisons there, similarly to what is already there (via %%time and %%memit).

Could also consider choosing optimal parameter combo in the other notebooks for SurfaceForcing, InitialConditions, and BoundaryForcing, but not elaborate further in these notebooks to not make them overly lengthy. Instead, we could add a note in the "Note" box that encourages the user to check out the "Using Dask" notebook for further explanation on use_dask. The added note could say something about that "Using Dask" also explains usage of "chunks" and "initial_slicing"

roms_tools/setup/boundary_forcing.py
blsaenz: Removed.  Sorry, this was left in when I went down a rabbit hole of translating lat/lon slices into ROMS slicing.

roms_tools/setup/surface_forcing.py
blsaenz: Fixed.

roms_tools/utils.py
blsaenz: Yes

roms_tools/utils.py
blsaenz: I replaced the boolean param -> unchunk_lateral_dims globally.  I think this better indicates what is happening.  I left unchunk_dask, so now the names correspond.

roms_tools/setup/boundary_forcing.py
blsaenz: Done

roms_tools/setup/initial_conditions.py
blsaenz: Done

roms_tools/setup/boundary_forcing.py
blsaenz: Done

roms_tools/setup/initial_conditions.py
blsaenz: That was leftover - thanks for the catch.  Removed, I believe nothing is defaulting to single chunks now.

roms_tools/setup/boundary_forcing.py
blsaenz: That was leftover - thanks for the catch. Removed, I believe nothing is defaulting to single chunks now.

roms_tools/setup/boundary_forcing.py
blsaenz: Removed

roms_tools/setup/surface_forcing.py
blsaenz: Reworded

roms_tools/utils.py
NoraLoose: So the chunk size won't be determined by the data, as it says in the docstrings. ;)

roms_tools/utils.py
NoraLoose: but we also have the parameter "reset_chunking" somewhere, which basically does the same thing?

roms_tools/datasets/lat_lon_datasets.py
```diff
@@ -1630,6 +1666,10 @@ def match_subdomain(self, target_coords: dict[str, Any]) -> None:
             A dictionary containing the target latitude and longitude values to select.
             Expected keys: "lat" and "lon", each mapped to a DataArray of coordinates.
 
+        reset_chunking : bool
```
NoraLoose: Here we use the terminology "reset_chunking". Should we say "unchunk_dask" instead, to be consistent?

roms_tools/setup/boundary_forcing.py
NoraLoose: Maybe changes are not pushed yet?

roms_tools/setup/boundary_forcing.py
NoraLoose: Not pushed yet, though?

roms_tools/setup/boundary_forcing.py
NoraLoose: Not pushed yet?

roms_tools/setup/boundary_forcing.py
blsaenz: removed/revised

blsaenz:
using_dask.ipynb updated with sections on chunks and initial_slice_bounds, and comments addressed I think.  I used Scott's dask preprocess refactoring - thanks.

roms_tools/datasets/lat_lon_datasets.py
blsaenz: Changed with global reset_chunking -> unchunk_lateral_dims

roms_tools/utils.py
blsaenz: Accepted!

blsaenz:
> Maybe you could add a new section there that explains what the new parameters do. Could also do some timing and memory profiling comparisons there, similarly to what is already there (via %%time and %%memit).

I would like to add the above to the docs already done, but I need to switch off this task soon.  We have the params now that could be tuned, but the default is now allowing processing larger datasets/grids and not bust memory.  

I think the first is maybe to limit the # of dask thread on large anvil core requests, but we'll see if that is still an issue.

ScottEilerman (approved):

PR #583, opened 2026-04-10
Updates to nesting

### LINKED ISSUE #581: Follow-up suggestions after nesting refactor
I just skimmed the docs after the nesting refactor. Nice work, @smaticka! I just have a few suggestions:

1. Docs: After creating the child grid in [this](https://roms-tools.readthedocs.io/en/latest/initial_conditions.html#Initializing-from-ROMS-Restart-Files), the align_grids function needs to be applied, before the initial conditions can be created.

2. For the plot_nesting function, I would make it easier on the user (the average user doesn't know about "straddle") and just change the syntax to plot_nesting(parent_grid, child_grid, verbose).

3. Nits for the nesting docs (but up to you):
  - I would define a dictionary with the child grid parameters, which are then passed into child_grid and child_grid_without_parent. That avoids repetition of the parameters and assures the user that the same parameter sets are used twice.
  - When calling the functions align_grids and make_edata in the docs, I would add clarity by writing make_edata(parent_grid=parent_grid, child_grid=child_grid, filename=filename, ...). That shows the user the explicit name of the keyword arguments.

smaticka:
This PR addresses all of the points in issue #581 

 Resolves #581

- [x] Tests fixed


roms_tools/plot.py
```diff
@@ -142,11 +142,14 @@ def plot_nesting(parent_grid_ds, child_grid_ds, parent_straddle, with_dim_names=
...
-    parent_lat_deg = parent_grid_ds["lat_rho"]
+    parent_lon_deg = parent_grid.ds["lon_rho"]
+    parent_lat_deg = parent_grid.ds["lat_rho"]
 
-    child_lon_deg = child_grid_ds["lon_rho"]
-    child_lat_deg = child_grid_ds["lat_rho"]
+    child_lon_deg = child_grid.ds["lon_rho"]
+    child_lat_deg = child_grid.ds["lat_rho"]
+
+    if parent_straddle is None:
```
ScottEilerman: Can we drop `parent_straddle` as an arg altogether and always use `parent_grid.straddle`? I'm not sure why someone would want to set it differently.

NoraLoose:
<p>Why does it take so much longer on anvil than on perlmutter? :/</p>

---

 <a href='https://app.reviewnb.com/CWorthy-ocean/roms-tools/pull/583/discussion/#comment-4239390440'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

smaticka:
@NoraLoose I was wondering the same thing. I ended up testing it a few days later and it was fine on Anvil, so I think it was probably just a slow Anvil day. I can rerun and update the notebook in ROMS-Tools it if we think it would be misleading for what users should expect.

NoraLoose:
Yep, I think rerunning is a good idea!

ScottEilerman:
It looks like the nesting notebook is still missing cell outputs for the last ~1/3 of the cells. Otherwise this is good to go!

roms_tools/plot.py
smaticka: Good point. I think it would always be the parent's `straddle`. If someone wanted the child's `straddle`, they probably set up their domain wrong. just put in the change.

smaticka:
I removed parent_straddle from plot_nesting, and reran the full notebook. the PR should be good to go now!

ScottEilerman (approved):

PR #585, opened 2026-04-14
Ensure longitude btwn 0 to 360

smaticka:
When ROMS-Tools creates a Grid, it makes sure the longitude is between 0 to 360 by adding 2*pi if any value is negative. This doesn't handle `center_lon` that are well out of the normal range (-720 or 720) for example. 

`straddle` is later determined based on the difference in longitude between two consecutive points and whether that difference exceeds 300 (true when it goes from 360 to 0), but this fails to register `True`, if the center longitude (`center_lon`) is either 720 or -720 (each equivalent to 0 degrees and should register `True` for `grid.straddle`). 

With the changes added, it will ensure any longitude set will be within the range of 360 degrees. 

### DIFF
--- a/roms_tools/setup/grid.py
+++ b/roms_tools/setup/grid.py
@@ -1058,8 +1058,7 @@ class Grid:
 
             # make sure lons are in [0, 360] range
             for lon in ["lon", "lonu", "lonv", "lonq"]:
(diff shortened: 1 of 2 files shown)


NoraLoose:
Can we add a test that would have caught the previous bug? :)

smaticka:
Yes! Definitely should do that.

smaticka:
ok, test is added!

ScottEilerman (approved):

Issue #450, opened 2025-09-03
Include 1 time step before mdl start to handle machine precision issues


smaticka:
Currently, if the forcing records start at the _same_ time as the model start, when roms interpolates the data to the start time timestamp, it will sometimes not find the first provided time step due to machine precision error. 

Adding an additional record (before the mdl start time) to all forcing files in roms-tools will help avoid this issue.

PR #586, opened 2026-04-15
Shift radiation time by 30 minutes

smaticka:
The data reported in ERA5 data is a cumulative over the time step (an accumulation from the previous time step to the current time). For this to be properly interpolated in ROMS (which takes 2 surrounding data points and interpolates to a given time point), the ERA5 data should be represented halfway between the time step (i.e. 30 minutes prior). 

This PR creates a new time dimension called `rad_time` that is equal to `time` minus 30 minutes. The `time` dimension for short and long wave radiation is then swapped to use `rad_time` instead. 

- [x] Test data fixed


ScottEilerman (commented):
Code seems fine. Looks like we'll need to save out new versions of the validation test data -- tests are failing because they are checking if we match previous results, which didn't have rad_time in the dims. @NoraLoose , is there a script anywhere for updating these, or do you just do it manually? I've started poking around but haven't found one yet.

ScottEilerman:
> Code seems fine. Looks like we'll need to save out new versions of the validation test data -- tests are failing because they are checking if we match previous results, which didn't have rad_time in the dims. @NoraLoose , is there a script anywhere for updating these, or do you just do it manually? I've started poking around but haven't found one yet.

Nevermind Nora, I just found it. Should have spent thirty more seconds before @'ing you!

@smaticka see here: https://github.com/CWorthy-ocean/roms-tools/blob/main/roms_tools/tests/test_setup/test_validation.py#L41

Did we decide if there are updates needed in our default ROMS configurations to include rad_time?

NoraLoose:
Two comments:

1. Shifting by 30 minutes is specific to the ERA5 dataset because the ERA5 variables are integrated over 1 hours in the source dataset, see [here](https://github.com/CWorthy-ocean/roms-tools/blob/00f3979dd1ce2da3fb08ee917ef000dd280444d4/roms_tools/datasets/lat_lon_datasets.py#L1529). But other source datases may be different. For example, I think JRA55 is a 3-hourly dataset. We therefore lose a bit of modularity if we hardcode the shift by 30 minutes in the `SurfaceForcing` class. I see two ways to fix this:
  - Introduce `rad_time` in the `ERA5Dataset.post_process` method. Probably the cleanest because here we also convert hourly integrated radiation values to fluxes. There may be downstream problems though if the `Dataset` class has two time dimensions, not sure.
  - Introduce `rad_time` in `SurfaceForcing._write_into_dataset` (this is also where the rest the time dim stuff is handled), but with a conditional that makes sure we are dealing with ERA5 as source. One minor issue with this approach (similar to the currently implemented in this PR) is that the radiation correction uses the unshifted time value, so the correction will be off by 30 minutes. But maybe okay since the correction is a climatology (12 monthly values), so  being 30 minutes off should be fine.

2. With the shifted `rad_time`, are we still sure that we meet the requirement that there is at least one value of `rad_time` before `start_time` and at least one value of `rad_time` after `end_time`? That's important, otherwise the model may crash (see for example https://github.com/CWorthy-ocean/roms-tools/issues/450). I don't think we are meeting this requirement now. The relevant place to check/change the code for this is this: https://github.com/CWorthy-ocean/roms-tools/blob/00f3979dd1ce2da3fb08ee917ef000dd280444d4/roms_tools/datasets/utils.py#L186

ScottEilerman (approved):
Approved once test data is updated!

smaticka:
Thanks for the review @NoraLoose ! All great points. 

I added a few fixes to the code to address your points. The test data has been updated too.

To handle the need to have 1 data point before the start and 1 after the end, we add 1 hour to the `end_time` before retrieving the data if it's ERA5. This provides 2 extra datapoint in the end of the time series. No adjustments are needed for the `start_time`. Two points at the end is needed if the `end_time` is between the 31-59 minutes mark (e.g. RT will pull the next hour, but then it will be shifted to the 30 minute mark prior and no data point will be available on the other side of the end_time). 

We ended up applying the shift in radiation time in the __post_init__ of SurfaceForcing but with a check that it is ERA5 type. If we allow for more types, this could be passed directly into the `_apply_rad_time` function and handled according to the timestep of each data source.

PR #587, opened 2026-04-17
zarr read chunking bug fix, dask documentation improvements

blsaenz:
This PR fixes a bug with zarr streaming (and likely on-disk zarr reading) where new default 'chunks' (that benefit unchunked netCDF file reading) were passed to dask-enabled zarr streaming/reads.  Dask normally will inherit efficient chunking from the underlying zarr chunking, but by imposing default chunking there were conflicting/non-aligned chunks and that broke streaming.  

Also, documentation errors were fixed and new documentation related to dask was edited for clarity.


docs/using_dask.ipynb
```diff
@@ -7071,15 +7071,11 @@
...
...g into pieces that are manageable with limited memory and processing.  Imposed default chunking is used because many large datasets used as source data by ROMS-Tools are unchunked netCDF files.  (Without imposed dask chunking, dask assumes all unchunked source data in a file should be fully read into memory, which is almost always undesirable).  For data sources that have their own intrinsic chunking (as is almost always the case with zarr datasets, and some netCDF datasets) it is better to leave those dimensions unchunked in dask, because dask will then inherit chunking from the data source. Using imposed dask chunks that do not align with underlying source data chunks often leads to large memory usage and very poor performance. Defaults chunks also typically use a chunk size of 1 on time-related dimensions, since many processing operations occur on single time slices.  In general, using smaller chunks avoids loading more data than needed, but with the possible penalty of slower data loading and computations. Using very large chunks or chunks that span the entire dataset dimension (e.g. {\"latitude\":-1}) can lead to excessive memory use and even crashes depending on the system. "
```
ScottEilerman:
```suggestion
    "The default for many datasets used my ROMS-Tools is to impose specific integer dimensions of 50 with geographic chunking (e.g. along latitude, longitude, eta_rho, eta_xi and other geographic dimensions). Dask requires chunking to break data processing into pieces that are manageable with limited memory and processing.  Imposed default chunking is used because many large datasets used as source data by ROMS-Tools are unchunked netCDF files.  (Without imposed dask chunking, dask assumes all unchunked source data in a file should be fully read into memory, which is almost always undesirable).  For data sources that have their own intrinsic chunking (as is almost always the case with zarr datasets, and some netCDF datasets) it is better to leave those dimensions unchunked in dask, because dask will then inherit chunking from the data source. Using imposed dask chunks that do not align with underlying source data chunks often leads to large memory usage and very poor performance. Default chunks also typically use a chunk size of 1 on time-related dimensions, since many processing operations occur on single time slices.  In general, using smaller chunks avoids loading more data than needed, but with the possible penalty of slower data loading and computations. Using very large chunks or chunks that span the entire dataset dimension (e.g. {\"latitude\":-1}) can lead to excessive memory use and even crashes depending on the system. "
```
typo: `Defaults` -> `Default`

ScottEilerman (approved):

docs/using_dask.ipynb
blsaenz: Accepted!

PR #588, opened 2026-04-18
Quick fix for faster zarr streaming - no chunks at all

blsaenz:
Testing zarr-backed ERA5 streamed surface forcing showed that passing None as chunks was significantly faster than passing {'time':1} or any more complex chunking.  With this PR, all xarray.open_zarr() calls are passed None as chunks, with a TODO to refactor chunking into the classes (which is non-trival since the ERA5 dataset class is used for different things using different source data files).  This fix is to enable large-domain, long run surface forcing as efficiently as possible.


blsaenz:
Sorry again, it seems that re-using a branch is a little confusing.  This PR is really only about the 'chunks=None for zarr is faster' commit.

ScottEilerman (approved):

PR #601, opened 2026-05-05
CDR time_interpolation

smaticka:
ROMS allows to either interpolate between prescribed cdr release points, or to treat it as a step function that holds a constant value until a new one is given. By default, ROMS does not use interpolation. 

This PR is to adjust ROMS-Tools to allow `time_interpolation` as an option. When set to true, linear interpolation is performed, and when it's False, the step function is used (this is set to default). It adjusts the way things are plotted (`plot_tracer_concentration`, `plot_volume_flux`) and how the total release is calculated (`compute_total_cdr_source`).

- [x] Add test to make sure `compute_total_cdr_source` is expected


smaticka:
@ScottEilerman, this should be ready for review!

roms_tools/tests/test_setup/test_cdr_release.py
```diff
@@ -408,13 +410,42 @@ def make_release(self, release_type, fluxes=None, concentrations=None, times=Non
...
             )
             tp._extend_to_endpoints(self.start, self.end)
             return tp
         else:
             raise ValueError(f"Unknown release type {release_type}")
 
-    @pytest.mark.parametrize("release_type", ["volume", "tracer"])
+    @pytest.mark.parametrize(
+        "release_type", ["volume", "tracer", "volume_interp_off", "tracer_interp_off"]
+    )
```
ScottEilerman: This all ends up working ok, though you can also double up on parameterize like

```python
    @pytest.mark.parametrize(
        "release_type", ["volume", "tracer"]
    )
    @pytest.mark.parametrize(
        "interp", [False, True]
    )
    def test_constant_flux(self, release_type, interp):
```

and it will do all combinations and pass the values separately so you don't have to parse later. I was a little surprised this worked to begin with, because `release_type` is a controlled value in parts of the backend (it can only be `volume` or `tracer_perturbation`), but it looks like this is a similar concept with the same name but not actually being used to control/alter/validate the same-named value on the backend.

roms_tools/setup/cdr_forcing.py
```diff
@@ -261,6 +261,7 @@ def _initialize_dataset(self, unique_times, unique_rel_times) -> xr.Dataset:
         ds["cdr_dep"] = ("ncdr", [r.depth for r in self.releases])
         ds["cdr_hsc"] = ("ncdr", [r.hsc for r in self.releases])
         ds["cdr_vsc"] = ("ncdr", [r.vsc for r in self.releases])
+        ds["cdr_interp"] = ("ncdr", [r.time_interpolation for r in self.releases])
```
ScottEilerman: I don't think ROMS actually uses this, does it? It just has a single flag in the `cdr_frc.opt` that controls all of the releases.

Which, I was going to comment later where we catch the "all releases must be the same" error later on in plotting -- we should catch that up-front rather than only when (if!) they plot. I'd have it mirror the `check_all_releases_same_type` pydantic validator on the `ReleaseCollector` object.

roms_tools/setup/cdr_forcing.py
```diff
@@ -439,14 +440,28 @@ def plot_volume_flux(
...
 
-        self._plot_line(
-            data,
-            release_names,
-            start,
-            end,
-            title="Volume flux of release(s)",
-            ylabel=r"m$^3$/s",
-        )
+        if all(self.ds["cdr_interp"]):
```
ScottEilerman: I guess we end up using that `cdr_interp` from the ds here, but I think it could/should just be a property on `ReleaseCollector` and/or `CDRForcing` since it doesn't need to go into the netcdf.

We can also remove the duplicated arguments between the two paths by doing

```python
    plotter = self._plot_line if self.releases.time_interpolation else self._plot_step
    plotter(data, release_names, start, end, title="Volume flux of release(s)", ylabel=r"m$^3$/s")
```

(assuming we implement `self.releases.time_interpolation`)

roms_tools/setup/cdr_forcing.py
```diff
@@ -505,14 +520,28 @@ def plot_tracer_concentration(
...
 
-        self._plot_line(
-            data,
-            release_names,
-            start,
-            end,
-            title=title,
-            ylabel=f"{self.ds['tracer_unit'].isel(ntracers=tracer_index).values.item()}",
-        )
+        if all(self.ds["cdr_interp"]):
```
ScottEilerman: same plot comments as before... though since we do this several times, we could also consider setting the `self._plotter` to either line/step once and for all in `_validate` (think of that like post-init) rather than re-do the if check I suggested in every plot method.

roms_tools/setup/cdr_forcing.py
smaticka: Ah, good point that it should catch it up front. I modeled it after the plot_volume_flux that throws the same error: https://github.com/CWorthy-ocean/roms-tools/blob/d613adf258ddcdd702aab4a73b4b8fd292b55eb6/roms_tools/setup/cdr_forcing.py#L432.

but yeah, it needs to do a proper check. I'll add a check in `ReleaseCollector` and  remove the ds["cdr_interp"] to rather just use `self.releases.time_interpolation` when needed.

roms_tools/tests/test_setup/test_cdr_release.py
smaticka: Ooh, that's a proper approach. thanks for the review! I'll fix that.

ScottEilerman (approved):

PR #604, opened 2026-05-11
Fix streaming bug

kthyng:

**Summary**

1. The chunking in `_load_from_copernicus()` in `lat_lon_datasets.py` was causing the Copernicus streaming to hang. I commented this out to avoid the issue since even a tiny grid hung meaning that any reasonably-sized grid would for sure hang too. The docs already steer users away from using streaming so I am not sure it is worth a lot of testing here to see if large domains with streaming from Copernicus would hit an issue specifically because of commenting out chunks with streaming vs. streaming with a large domain.
    1. An alternative solution would be to have the default be to download from Copernicus and then use that.
2. After that fix, the wrong day is grabbed for ini_time for `InitialConditions`  so I input `ini_time` to `end_time` for `_load_from_copernicuse()` if `end_time` is None which fixes it.

Addresses https://cworthy.atlassian.net/browse/CSD-872

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -31,6 +31,8 @@
 * Integrate narrow-channel closing directly into `Grid.update_mask()` (internal `_close_narrow_channels`), iterating north–south and east–west up to 10 passes
 * short and long wave radiation time is shifted 1/2 a timestep sooner and have a dim of `rad_time` ([#586](https://github.com/CWorthy-ocean/roms-tools/pull/586))
 * The coarse UNIFIED BGC dataset used for testing was updated to have depths of 0 and 5 m available ([#589](https://github.com/CWorthy-ocean/roms-tools/pull/589))
+* When using the default of streaming from Copernicus for GLORYS output, the domain is now subset by the bounds of the domain to save read time. ([#604](https://github.com/CWorthy-ocean/roms-tools/pull/604))
+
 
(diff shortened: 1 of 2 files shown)


kthyng:
This is meant to address [CI / 3.13-build-with-streaming](https://github.com/CWorthy-ocean/roms-tools/actions/runs/25697345624/job/75448797612?pr=604) specifically but it isn't run on PRs from forks. It will only be run once the branch is merged to main.

I guess we can still use this for discussion and hope/assume that it will work the same in CI as I have seen when running it locally for this test.

blsaenz (approved):
Hi Kristen - consider adding a code comment above the commented-out chunking saying something like 'The copernicus marine datasets have intrinsic chunking, which should be considered before any additional external chunking." or something like that?

kthyng:
@blsaenz I added some info to the function description summarizing how it works:

>       This reads in using zarr on the Copernicus Marine Data Store.
>       The impact of chunk_size_limit=-1 is to let the toolkit determine optimal chunking
>       based on the dataset and request parameters, which can help improve performance
>       when working with large datasets. So, this is dask-backed but does not use the
>       parameter _DEFAULT_LAT_LON_LATERAL_CHUNK for chunking, which does not work well
>       for streaming.

PR #608, opened 2026-05-20
Add MBL co2 data

smaticka:
Currently ROMS-Tools makes a surface forcing for atmospheric co2 using either the Unified bgc or CESM dataset. Both of which are climatology. This PR is designed to:

- add the option to use time-vaying CO2 data, taken from the NOAA GML, Marine Boundary Layer dataset
- remove the option to use Unified and CESM pco2 data

**Note,** roms will, by default, use a constant value for xco2 (PR needed in ROMS), and a cppdefs flag needs to be turned on to use time-varying. 
**Note,** a new SurfaceForcing object is needed to be made for the time-varying co2, in addition to the bgc climatology object. 

**Depends on:**
- roms-tools-test-data PR:  https://github.com/CWorthy-ocean/roms-tools-test-data/pull/32
- ucla-roms PR:  https://github.com/CWorthy-ocean/ucla-roms/pull/262

**To-Dos:**
- [x] Tests added
- [x] make PR in roms to add an if statement to use a constant xco2 by default. previously it looked for pco2
- [x] change URL back to main branch for downloading the test data.


smaticka:
To test this PR, a time-varying xco2 file was made for the GoM domain and two simulations were run - one with the time-varying xco2, and one with the constant xco2. The resulting CO2 fluxes from the model are compared to the fluxes that we'd expect to get based on the forcing applied. 

**To estimate the flux that we'd expect, a 'back of the envelope' approach was used. The following equations were used:**
`Flux = k * s * delta(pCO2)`. 
`k` is rate of gas transfer; `k = 0.251 * U^2 /100/3600` (U is wind speed at 10 meters. it's taken from the forcing files used for the sims); the time-average of `k` was used; `k` units are `m/s`. 
`s` is the solubility of CO2. It was estimated as a constant `29 mol / m^3 / atm` which is typical for tropical waters of ~25 Celsius; GoM was 20-25 C for the sim.
`delta(pCO2)` is the difference between the atmospheric and oceanic partial pressure of CO2; `delta(pCO2) = pCO2_atm - pCO2_ocean`; `pCO2_atm` is taken as the time-avg xco2 for the t-varying sim, and as `284.7` for the t-constant sime. `pCO2_ocean` is taken from the surface of the model and left as time-varying. 

These are the 1-month t-averaged flux results from the model:
image

These are the fluxes based on the equations above, 2/3rd through the simulation for a snapshot, using the same colorbar limits:
image

The spatial patterns and magnitude line up really well. The xCO2 is lower in the atm for the t-constant sim, leading to a spatial majority of off-gassing.

kthyng:
@smaticka I am going through this and have some questions to start. I ran tests locally once I got set up with this PR and associated links above, but I am hitting some failed tests. Maybe I don't have everything set up properly? I'm getting:

```
FAILED roms_tools/tests/test_setup/test_validation.py::test_check_results[bgc_surface_forcing] - AssertionError: Left and right Dataset objects are not close
FAILED roms_tools/tests/test_setup/test_validation.py::test_check_results[bgc_surface_forcing_from_climatology] - AssertionError: Left and right Dataset objects are not close
FAILED roms_tools/tests/test_setup/test_validation.py::test_check_results[bgc_surface_forcing_from_unified_climatology] - AssertionError: Left and right Dataset objects are not close
```

I also updated a time variable in `test_surface_forcing.py` from "xco2_air_time" to "xco2_time" in case that gives any hints.

kthyng:
Ah, do you still need to update the files that are being compared with? I bet that is it.

kthyng:
Also a question: we don't allow people to run CO2 as a climatology anymore? Maybe it would be useful to be specific about that, if so.

smaticka:
Thanks for the review @kthyng !

Good catch on the 2 typos ('helf' and 'xco2_air_time'). I fixed those. 
And yes, you're right on the data not being updated for `test_validate.py`. I updated those so it should work now. I must not have run a pytest on all of the test to catch that myself.

smaticka:
> Also a question: we don't allow people to run CO2 as a climatology anymore? Maybe it would be useful to be specific about that, if so.

@kthyng, Correct that it's no longer available. I'm more in favor of maintaining the release notes with what is no longer available: https://roms-tools.readthedocs.io/en/latest/releases.html

That way the current documentation contains what is available in ROMS-Tools, and if a user isn't seeing something they've used before, they can check the release notes. Otherwise, I could see there being a lot of notes piling up in the documentation about functionality we don't support anymore.. What do you think? Another option could be to have a temporary note that we take down, but then we need to remember to remove it at some point.

kthyng:
>  I'm more in favor of maintaining the release notes with what is no longer available

Yes that is a good way to go.

Sorry I meant to push my fix to that test but forgot to do so after I wrote the comments!

I am going to try a fix for those remaining tests that fail.

smaticka:
Nice, thanks for the fix! Looks like they're passing now.

docs/releases.md
```diff
@@ -3,11 +3,13 @@
 ## v4.0.0 (unreleased)
 
 ### Breaking Changes
+* pco2 is no longer read in during `SurfaceForcing` `type='bgc'` ([#608](https://github.com/CWorthy-ocean/roms-tools/pull/608)).
```
kthyng, line 6: One more comment but up to what you think is best: Would it help to add here that pco2 does not work with the clim=True option anymore in favor of the new time-varying approach? I can't remember exactly all the details of what a user is meant to choose as they go through, but it seems like this could be a source of confusion if they are meant to choose certain variables as clim=True and others for time-varying.

smaticka: We can add that if you think it would be helpful. I just tested running SurfaceForcing for co2 and set climatology to true to see how it errors out. It throws this error:
```
ValueError: The dataset contains 2209 time steps, but the climatology flag is set to True, which requires exactly 12 time steps.
```
We could add a check that if "name" is set to "MBL_co2", then check that clim is false. If it's not, give an error saying "When 'name' is 'MBL_co2', time-varying xco2 data is expected. `climatology` must be 'False'".

Do you think a check would be more helpful or do you think the error being thrown is sufficient? 
If the current error is sufficient, I'll add a small note to the release notes. 

I guess my thoughts are that it's a 'Time-varying CO2' section in the docs, that specifies the allowable data source, so it's implicit that climatology is not used. I'm ok with the current approach, but open to other thoughts!

kthyng: I do think that check would be helpful. More what I am thinking about is the why behind the change. Maybe better than the release notes would be in the "Time-varying CO2" section itself, explaining why we have CO2 as time varying instead of climatological. I assume it is something like: it is more realistic and we are able to get reasonable data for it, so we make that the available option beyond just using a constant.

smaticka: ok, I just added the check. 

I see. Maybe what's missing is more broadly explaining to the user when climatolgical data is or isn't appropriate. I could add a side note explaining climatology in general and why we offer xco2 time-varying. 

If someone wanted to use ROMS MARBL for non CDR scenarios, I could see climatological xco2 being ok as long as it represented the period being modeled accurately, and they didn't want the sub-monthly variations. So I'm not sure if we won't ever allow it?

kthyng: Yeah I think it is fine to keep the note limited to the reasoning behind the current options, for now.

kthyng (approved):
Looks great!

PR #611, opened 2026-05-30
Adding create_movie() in ROMSOutput class + support in plot.py

blsaenz:
This PR adds movie-making abilities to ROMSoutput (ROMSoutput.create_movie()), with similar inputs as the plot() method. It builds upon the plot() functionality in plot.py, but required a bit of abstracting and refactoring to re-use that code.  

Example:

cdr_output.create_movie(
    "ALK_source",
    time_range=slice(0, -1, 6),                # every 6 output (here, every six hours)
    fps=6,                                     # frames per second
    output_file="simulation_ALK_source.mp4",
    s=-1,                                                      # surface
    include_boundary=True,                     # include boundary grid cells
    timestamp_xy=(0.05, 0.95)                  # display a time stamp at plot axes (x,y)
)

- [X] Tests added
- [X] Passes `pre-commit run --all-files`


blsaenz:
Adding PR for visibility.  I still need to create documentation.

roms_tools/analysis/roms_output.py
```diff
@@ -216,6 +225,191 @@ def plot(
             cmap_name=cmap_name,
         )
 
+    @staticmethod
+    def _cmap_name_for_var(var_name: str) -> str:
```
kthyng, line 229: I have a package called xcmocean that I haven't looked at for awhile (https://github.com/pangeo-data/xcmocean) which is meant to do this job in a more general way: use regex to match variable names to the likely intended cmocean colormap. It's a bit kludgey with the variable matching, but has some good parts.

roms_tools/analysis/roms_output.py
```diff
@@ -216,6 +225,191 @@ def plot(
...
+
+    def _load_field_slice(self, field: xr.DataArray) -> xr.DataArray:
+        if self.use_dask:
+            from dask.diagnostics import ProgressBar
+
+            with ProgressBar():
+                field.load()
+        return field
+
+    def create_movie(
```
kthyng, line 259: does this also save the individual images that are being used to create the animation? I often like to use those individually too (then link with ffmpeg), personally, so it might be nice if that is possible with this approach.

kthyng (commented):
Is it a good time to try this or are there more changes coming to wait for? I haven't tried it yet but would like to!

roms_tools/analysis/roms_output.py
blsaenz: No, in fact I used the Animation class in matplotlib specifically to avoid creating gigabytes of individual plots.  I think it's cleanest just to give what is is the name - a single movie .  And then if you find a frame where an individual plot is needed, it is easy to call plot() with the same parameters to generate it the time/slice where you need it?

roms_tools/analysis/roms_output.py
blsaenz: We can consider it but maybe in a differenct PR?  This function is actually just a re-wrapping of what ROMS-Tools does now for plot() for picking a colormap, so I am not introducing anything new or exciting here.  I don't even know if we can use a custom colormap. Design-wise, I am under the impression we are not trying to support 'fancy' plotting options, but it would b good to bring up in standup or backlog grooming since I don't know for sure.

blsaenz:
Updated documentation!  Ready for full review I think

roms_tools/analysis/roms_output.py
kthyng: Yeah fair enough!

kthyng (approved):
This looks reasonable.

smaticka (approved):
looks good to me! Thanks for the new function @blsaenz . I changed the paths in the plotting notebook to anvil, and added a player for the mp4. 

I think if we update this file to include pip installing with `movie`, it should be good to go!
`/anvil/projects/x-ees250129/x-smaticka/roms-tools/docs/installation.rst`

PR #615, opened 2026-06-09
Bgc riverinput from file

ubbu36:
## Summary

Adds RIVR2O-based river BGC forcing to `RiverForcing`, alongside the existing Dai/Trenberth discharge workflow. River mouths can receive spatially and temporally varying carbon and nutrient export from annual RIVR2O files, while discharge can remain a repeating monthly climatology when Dai data are missing.

Also moves MARBL river tracer defaults from hardcoded values to `river_tracer_defaults.nc` (32 MARBL tracers plus `temp` and `salt`, 34 total), downloaded via the same `river_data` pooch registry as Dai discharge.

Refactors river BGC so `RiverForcing` is data-source agnostic: RIVR2O export conversion, MARBL tracer mapping, and gap-filling are handled by BGC dataset classes and a shared merge utility, not hardcoded in the forcing orchestrator.

## Motivation

Users should be able to use observationally constrained global river BGC export (RIVR2O) instead of uniform default concentrations, including the common case of repeating seasonal discharge (Dai climatology) combined with year-to-year BGC variation from RIVR2O annual files.

River BGC products are patchy by nature. This PR separates **discharge** (`source` → Dai) from **BGC** (`bgc_source`), and uses an explicit **fill** policy for tracers the dynamic source does not provide, rather than bespoke merge logic inside `RiverForcing`.

### RiverForcing (orchestration)

- New `bgc_source` options: `CONSTANTS` (default) or `RIVR2O` (requires `path`).
- Optional `bgc_source["fill"]` (default `{"name": "CONSTANTS"}`) supplies scalar concentrations for tracers missing from the dynamic source or non-finite at a given `(river_time, nriver)`.
- Unified BGC path for all sources: `_apply_bgc_tracers()` calls `forcing_concentrations()` on the configured dataset, then merges with fill via `fill_river_bgc_concentrations()`.
- `RiverForcing` no longer contains RIVR2O unit conversion, stoichiometry, or MARBL tracer name logic.
- When Dai discharge is climatology and BGC requires a calendar axis (`requires_calendar_discharge_time`), expands `river_time` to a monthly calendar axis; `river_volume` repeats by month while BGC varies by year via `abs_time`.
- Sets `discharge_climatology` dataset attribute for plotting and time-axis behavior.

### BGC dataset layer (`river_datasets.py`)

- **`RiverBGCDataset` protocol**: `forcing_concentrations()` and `requires_calendar_discharge_time`.
- **`RiverTracerDefaultsDataset`**: loads recommended defaults from NetCDF; implements `forcing_concentrations()` for constant fields.
- **`Rivr2oRiverBGCDataset`**: owns RIVR2O I/O, DIC-cell sampling, discharge partitioning, export→concentration conversion, and MARBL tracer combinations via `forcing_concentrations()`.
- **`fill_river_bgc_concentrations()`**: generic merge of dynamic `(river_time, nriver)` arrays with scalar fill values.

### Rivr2oRiverBGCDataset (RIVR2O product)

- Loads and concatenates yearly `rivr2o_riverinputs_*.nc` files.
- Renames DIN→NO3 and DIP→PO4; keeps carbon pools separate.
- Uses nearest grid cell with positive DIC export (DIC-only mask).
- `discharge_partition_weights` splits export among rivers sharing a cell in proportion to discharge, normalized over `river_time` so co-located rivers get similar concentrations.
- Warns if sampled DIC cell is more than 100 km from the river mouth.
- Year-based time subsetting for July mid-year timestamps; clamps to 1903–2024.

### Tracer defaults and downloads

- **`get_tracer_defaults()`** reads recommended values from `river_tracer_defaults.nc`.
- **`download_river_tracer_defaults()`** aligned with other downloads: registered on the `river_data` pooch object (same pattern as `download_river_data()`), no separate pooch registry or bundled fallback.

## Docs and tests

- Updated `docs/river_forcing.ipynb` with RIVR2O examples and `bgc_source["fill"]` documentation.
- Updated `docs/methods.md` River Forcing section for optional BGC and fill.
- New and updated unit tests for `fill_river_bgc_concentrations`, RIVR2O dataset logic, `RiverForcing` BGC sources, tracer defaults, and validation snapshots (`river_forcing_with_bgc`, `river_forcing_no_climatology`).
- Changes noted in `docs/releases.md`.
- Passes `pre-commit run --all-files`.


roms_tools/datasets/download.py
blsaenz, line 1: The functionality in download seems like it is doing something different from all other downloads, coming from the same repos, and I don't understand why.

download_river_tracer_defaults() here doesn't seem to look like the other downloads.  Can you make it download_river_tracer_defaults_data(), and match the inputs/outputs and structure to the other download_*_data() methods?

roms_tools/setup/utils.py
blsaenz, line 1: Here we have added lru_cache. What is it doing and why do we need it, compared to the other methods in download?

roms_tools/setup/river_forcing.py
```diff
@@ -77,6 +95,38 @@ class RiverForcing:
...
+        For ``"RIVR2O"``, the product covers 1903-2024. Simulation times before 1903
+        or after 2024 use the 1903 and 2024 fields, respectively. Each river mouth is
+        mapped to the nearest RIVR2O grid cell with positive ``DIC`` export; that cell
+        supplies all tracers, with exports interpolated in time onto ``river_time``.
+        If several rivers share one RIVR2O cell, export is split in proportion to
+        discharge each month and normalized over the year so co-located rivers keep
+        similar concentrations in every month while higher-discharge months carry
+        more carbon export.
+
+        When Dai discharge is monthly climatology (``convert_to_climatology`` is
```
blsaenz: Can you rewrite this comment section?  I can't understand, especially the last part.

roms_tools/setup/river_forcing.py
```diff
@@ -253,6 +322,263 @@ def _get_data(self):
 
         return data
 
+    def _get_bgc_data(self) -> Rivr2oRiverBGCDataset:
```
blsaenz: This section is very specific to the underlying dataset, and should be refactored into the matching dataset.  Now that we have more than one option, I think this class needs a refactor to be dataset agnostic, similar to the other setup classes.

roms_tools/setup/river_forcing.py
blsaenz: Still trying to understand this.  On the basis of the compilation of this docstring, I am worried that we are assuming too much.  I feel like it should be obvious that the user is either using/requesting a climatology for either flows or BGC, and then that's what you get out the other side.  Either way, I'm not sure having this detail here is helping elucidate what is happening.

roms_tools/setup/river_forcing.py
```diff
@@ -77,6 +106,43 @@ class RiverForcing:
...
+
+        For ``"RIVR2O"``, the product covers 1903-2024. Dynamic tracers are merged
+        with fill values for all other BGC tracers in the ROMS schema. Simulation times before 1903
+        or after 2024 use the 1903 and 2024 fields, respectively. Each river mouth is
+        mapped to the nearest RIVR2O grid cell with positive ``DIC`` export; that cell
+        supplies all tracers, with exports interpolated in time onto ``river_time``.
+        If several rivers share one RIVR2O cell, export is split in proportion to
+        discharge each month and normalized over the year so co-located rivers keep
+        similar concentrations in every month while higher-discharge months carry
+        more carbon export.
```
blsaenz: All of the details of particular BGC datasets, I think, should be in the datasets themselves.  Here in RiverForcing, we should just be describing how we use one to fill another if necessary, and not the details about which variables are present in which datasets

roms_tools/setup/river_forcing.py
```diff
@@ -253,6 +349,160 @@ def _get_data(self):
...
+            lats.append(np.mean(cell_lats))
+        return np.asarray(lons), np.asarray(lats)
+
+    def _set_river_tracer_values(
+        self, ds: xr.Dataset, tracer_name: str, values: xr.DataArray
+    ) -> None:
+        ntracer = int(np.where(ds.tracer_name.values == tracer_name)[0][0])
+        ds["river_tracer"].loc[{"ntracers": ntracer}] = values
+
+    @staticmethod
```
blsaenz: This could move to utils.py and get imported

roms_tools/setup/river_forcing.py
```diff
@@ -253,6 +349,160 @@ def _get_data(self):
...
+                dates.append(candidate)
+            month += 1
+            if month > 12:
+                month = 1
+                year += 1
+        if not dates:
+            raise ValueError("No mid-month dates fall between start_time and end_time.")
+        return dates
+
+    def _expand_climatology_for_rivr2o_timeseries(self, ds: xr.Dataset) -> xr.Dataset:
```
blsaenz: Can we generalize this away from rivr2o?  It seems like we may have both cases of flow climatology + time series BGC, and time series flow + climatology BGC that this code could be used for?

roms_tools/datasets/river_datasets.py
```diff
@@ -1,19 +1,280 @@
...
+                    np.float32
+                )
+            else:
+                merged[tracer_name] = dyn.astype(np.float32)
+        else:
+            merged[tracer_name] = fill_arr
+    return merged
+
+
+EXPECTED_TRACER_NAMES = (
```
blsaenz: Consider changing name to MARBL_TRACER_NAMES, move to utils.py and import?

roms_tools/datasets/river_datasets.py
```diff
@@ -443,6 +708,587 @@ def add_time_info(self, ds: xr.Dataset) -> xr.Dataset:
...
+                lon_dim: xr.DataArray(nearest_lon, dims="points"),
+            }
+        )
+
+        if scalar_point:
+            return sampled.squeeze("points", drop=True)
+
+        return sampled
+
+    def nearest_dic_cell_indices_for_points(
```
blsaenz, line 1044: This will be slow for 10,000 rivers.  Let's leave as-is for now and speed it up when we merge with Abigale's code.

roms_tools/setup/utils.py
```diff
@@ -674,50 +679,22 @@ def add_tracer_metadata_to_ds(ds, include_bgc=True, with_flux_units=False):
 def get_tracer_defaults() -> dict[str, float]:
     """Returns constant default tracer concentrations for ROMS-MARBL.
 
-    These values represent typical physical and biogeochemical tracer levels
-    (e.g., temperature, salinity, nutrients, carbon) in freshwater.
+    Values are read from ``river_tracer_defaults.nc`` (recommended values at
```
blsaenz, line 818: Maybe explain in the docstrings of the function that these things live here due to a circular import issues and for re-use across multiple RiverForcing datasets and related calls.  It took me a while to figure out why this was here and what was going on.

blsaenz (changes requested):
Almost there!  There is only one new comment of substance I think.  Also, could you respond directly to each of the comments so we have records of what happened?  Can be as sort as 'implemented' or 'done,'  but if things were moved it is good to say where, how things changed etc.

roms_tools/setup/utils.py
blsaenz: NVM, I looked at it more, seems OK

roms_tools/tests/test_datasets/test_river_datasets.py
abigale-wyatt, line 1: Do we need to add a test dataset to here: [roms-tools-test-data](https://github.com/CWorthy-ocean/roms-tools-test-data) ? 

I genuinely don't know when we make them and what are good tests.

roms_tools/setup/river_forcing.py
```diff
@@ -253,6 +349,160 @@ def _get_data(self):
...
+        ds_expanded.attrs[DISCHARGE_CLIMATOLOGY_ATTR] = "True"
+        ds_expanded, time = add_time_info_to_ds(
+            ds_expanded,
+            self.model_reference_date,
+            climatology=False,
+            time_name="river_time",
+        )
+        logging.info(
+            "Expanded discharge climatology to %d calendar river_time steps "
+            "(%s to %s) for RIVR2O BGC.",
```
abigale-wyatt: I didn't understand what this INFO meant at first. Maybe we can come up with something a little more clear like:
`f"Tiled climatological river discharge in time for {n_steps} monthly river_time steps to apply RIVR2O BGC."`

roms_tools/datasets/river_datasets.py
```diff
@@ -443,6 +708,587 @@ def add_time_info(self, ds: xr.Dataset) -> xr.Dataset:
...
+        straddle: bool = False,
+        method: str = "nearest",
+    ) -> xr.Dataset:
+        """Sample MARBL tracer exports at point locations.
+
+        For each query point, the nearest grid cell with positive ``DIC`` export
+        is used for **all tracers and all years** in the loaded dataset. Cell
+        choice does not depend on simulation year; exports vary in time only
+        along the ``time`` dimension of the returned sample.
+
```
abigale-wyatt, line 997: I saw the comment at the top that DIC had a consistent mask year to year, but DIP has better coverage. Probably fine to use DIC for now as it means we only need to check one cell, but it means that there are times when there may be PO4 data closer to the river mouth than the cell we're using (which has DIC).

roms_tools/datasets/river_datasets.py
```diff
@@ -443,6 +708,587 @@ def add_time_info(self, ds: xr.Dataset) -> xr.Dataset:
...
+        )
+
+    def clean_up(self, ds: xr.Dataset) -> xr.Dataset:
+        """Replace fill values and rename raw RIVR2O variables."""
+        lat_dim = self.dim_names["latitude"]
+        lon_dim = self.dim_names["longitude"]
+
+        for var_name in self.var_names.values():
+            if var_name in ds:
+                ds[var_name] = ds[var_name].where(ds[var_name] != self.fill_value)
```
abigale-wyatt, line 854: There are some negative values for tracers in the rivr2o dataset that Yangyang masked out when he masked the fill values as well. I think we should do the same.

roms_tools/datasets/river_datasets.py
```diff
@@ -443,6 +708,587 @@ def add_time_info(self, ds: xr.Dataset) -> xr.Dataset:
...
+                    label,
+                    min_dist_m / 1000.0,
+                    q_lat,
+                    q_lon,
+                    int(lat_indices[nearest]),
+                    int(lon_indices[nearest]),
+                )
+            nearest_lat.append(int(lat_indices[nearest]))
+            nearest_lon.append(int(lon_indices[nearest]))
+        return nearest_lat, nearest_lon
```
abigale-wyatt, line 1090: I don't think the nearest neighbor approach is working well enough... I had Claude check for Iceland and if you sum all the nutrients loads round the whole island like > 70% of the nutrient load isn't being delivered to roms by the D&T rivers and the river forcing produced here. 

The scripts I gave had some accounting that used a 3deg box and then divided the nutrient load amongst the rivers in order to get concentrations that would make sure the global nutrient budget closes.

I don't know if we need to close a domain budget (though maybe that's the right approach?), but I think at least aggregating all RIVR2O cells within a search window around each river mouth is potentially a reasonable approach. 3 deg is too big for iceland, but maybe you can use someting that checks x number  of nearby of grid cells?

roms_tools/datasets/river_datasets.py
```diff
@@ -1,19 +1,280 @@
...
+    fill_at_nan: bool = True,
+) -> dict[str, xr.DataArray]:
+    """Merge dynamic river BGC concentrations with fill values.
+
+    For each tracer in ``tracer_names``:
+
+    - If the tracer is absent from ``dynamic``, broadcast the fill scalar to
+      ``template`` shape.
+    - If the tracer is present and ``fill_at_nan`` is True, use dynamic values
+      where finite and fill elsewhere.
```
abigale-wyatt, line 85: I think this could lead to strange values? e.g. a river has generally low NO3 compared to the global average, but the record is missing a year.... 

So I guess think if the dynamic values are inf or nan at some time steps but not others, that a linear interpolation makes more sense than filling with the constant global values.

roms_tools/setup/river_forcing.py
abigale-wyatt: Also would be useful to have a generalized time step. GloFAS has daily data that we want to work with the rivr2o BGC.

abigale-wyatt (commented):
One major comment is that I think the nearest neighbor approach isn't working well enough here. I think something more similar to the aggregated box in the matlab script from Princeton would help. Let me know if you want to discuss? 

I also added some comments to the notebook.

roms_tools/datasets/download.py
ubbu36: done

roms_tools/setup/utils.py
ubbu36: ok

roms_tools/setup/river_forcing.py
ubbu36: done

roms_tools/setup/river_forcing.py
ubbu36: Split up into smaller building blocks, with more general names.

roms_tools/datasets/river_datasets.py
ubbu36: pause for now, and review after incorporating new river discharge data.

roms_tools/datasets/river_datasets.py
ubbu36: I landed on this because I didn't want to draw PO4 from a different cell than DIC, so will leave it in for now.

roms_tools/setup/river_forcing.py
ubbu36: I made the INFO log clearer now, I hope.

roms_tools/datasets/river_datasets.py
ubbu36: I agree. I've added a function that interpolate temporally if a spot is missing data for one year (but has at least two values in the time period chosen). Hopefully, I've implemented it to be data agnostic following the general philosphy.

roms_tools/datasets/river_datasets.py
ubbu36: I added _mask_invalid_dynamic_bgc_concentrations in river_forcing.py

roms_tools/setup/river_forcing.py
ubbu36: the BGC creation is already agnostic to the discharge timestep.

roms_tools/setup/river_forcing.py
ubbu36: hopefully it's better now.

roms_tools/setup/river_forcing.py
ubbu36: done.

roms_tools/datasets/river_datasets.py
ubbu36: done

roms_tools/setup/utils.py
ubbu36: I added an explanation.

roms_tools/tests/test_datasets/test_river_datasets.py
ubbu36: I added test data plus tests to make the rivr2o test-suite comparable with GLORYS, ERA5, etc.

roms_tools/setup/river_forcing.py
```diff
@@ -372,16 +586,18 @@ def _create_river_forcing(self, data):
         ds["river_volume"].attrs["long_name"] = "River volume flux"
         ds["river_volume"].attrs["units"] = "m^3/s"
 
-        # River tracers
-        ds["river_tracer"] = xr.zeros_like(
-            ds.river_time.astype(np.float32) * ds.ntracers * ds.nriver, dtype=np.float32
-        )
-        ds["river_tracer"].attrs["long_name"] = "River tracer data"
+        # River tracers (per-tracer units live on tracer_unit, not on this array)
```
blsaenz, line 623: For long times or large numbers of rivers, this code is potentially inefficient - we are creating a massive array just to use as a template for another massive array (unless we have loaded with dask, but I'm not sure how that would be effective with the individual rivers here).  It would be better to just find the dimensions and initialize the array with those integers rather than use zeros_like(): 

```
# Extract the target shape from the dataset's sizes
target_shape = (ds.sizes["river_time"], ds.sizes["ntracers"], ds.sizes["nriver"])
ds["river_tracer"] = xr.DataArray(
    np.zeros(target_shape, dtype=np.float32), 
    dims=("river_time", "ntracers", "nriver")
)
ds["river_tracer"].attrs = {"long_name": "River tracer data"}
```

roms_tools/tests/test_setup/test_utils.py
```diff
@@ -295,3 +300,127 @@ def test_check_and_set_full_user_boundaries(simple_mask):
...
+        dic = xr.DataArray(
+            values,
+            dims=["river_time", "nriver"],
+            coords={"river_time": abs_time, "nriver": [0]},
+        )
+        result = interpolate_dynamic_bgc_by_calendar_year({"DIC": dic}, abs_time)
+        gap = result["DIC"].sel(river_time=abs_time.dt.year == 2001)
+        np.testing.assert_allclose(gap.values, 150.0, rtol=1e-6)
+
+    def test_leading_and_trailing_nan_years_remain_nan(self):
```
blsaenz, line 400: Does this test indicate that the normal behavior is to not extrapolate past the end of river BGC data, but then above interior years are interpolated? Does that mean we leave Nans on the ends?

blsaenz (changes requested):
Maybe this will save specific comments.

roms_tools/tests/test_setup/test_utils.py
ubbu36: There is the case of 2024 and 1903, where we just repeat the value into the past and future.

But, this is the case of missing data in a given grid cell for a given year-span. In that case, it does not extrapolate data from the middle of the time-series into the end-years, and that's what the test asserts.

roms_tools/setup/river_forcing.py
ubbu36: is it better now?

ubbu36:
<p>I don't really like them, but I think that belongs in the total overhaul of the docs after we have the dicharge data updated.</p>

---

 <a href='https://app.reviewnb.com/CWorthy-ocean/roms-tools/pull/615/discussion/#comment-4781729066'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

blsaenz (approved):
Approve!

Issue #368, opened 2025-07-01
Out of memory when creating boundary forcing with 2D fill


abigale-wyatt:
When creating the boundary forcing for the PACMED domain, I run out of memory when using the 2D fill. In a jupyter notebook it crashes the kernel.

PR #616, opened 2026-06-09
Changes to speed up BoundaryForcing calls

kthyng:
This PR introduces a source-fill and regrid-engine API for `BoundaryForcing`, and deprecates the `apply_2d_horizontal_fill` boolean flag. Users can now explicitly control source-side filling (prefill: None (default), "2d_lateral_fill", "inverse_dist", "nearest_s2d", "nearest_neighbor") and choose the horizontal regrid engine independently (regrid_method: "auto" (default = xESMF if installed, else scipy), "xesmf", "scipy"). The default path is xESMF masked-bilinear interpolation with inverse-distance-weighted destination extrapolation when available, with a nearest-neighbor source pre-fill + scipy interp fallback on Windows/pip installs. When a prefill is set, the source is pre-filled so that nan's are not present (matching the legacy AMG precedent) before per-boundary regridding with plain bilinear, making extrap_method irrelevant. The deprecated apply_2d_horizontal_fill=True/False still works with a clear DeprecationWarning, mapping to the equivalent prefill values for backward compatibility.

Scope and testing: Changes are confined to BoundaryForcing only; InitialConditions, SurfaceForcing, Topography, and TidalForcing are intentionally unchanged and produce numerically equivalent output to pre-change. Shared validation helpers (validate_prefill, validate_extrap, resolve_regrid_engine) and an apply_prefill dispatcher on the dataset class enable forward-compatible rollout to other classes later. Comprehensive test coverage includes deprecation mapping, conflict detection, xESMF-availability gating, YAML round-trip compat (old and new formats), per-prefill-method NaN-free assertions, and cross-engine tolerances. All boundary/fill/prefill tests pass; 135 untouched-class tests pass. The physics boundary forcing regression fixture reflects the new xESMF default; BGC fixtures are pinned to regrid_method="scipy" for byte-reproducibility with pre-v4 outputs.

- [x] Closes https://github.com/CWorthy-ocean/roms-tools/issues/368
- [x] Tests added


kthyng:
Here are the timing differences for different approaches: 
image

roms_tools/datasets/lat_lon_datasets.py
```diff
@@ -709,6 +772,142 @@ def apply_lateral_fill(self):
...
+        lon = self.dim_names["longitude"]
+        lat_vals = self.ds[lat].values
+        lon_vals = self.ds[lon].values
+
+        def _build(mask_da):
+            ds_in = xr.Dataset(
+                {
+                    "mask": (
+                        ("lat", "lon"),
+                        mask_da.transpose(lat, lon).values.astype("int32"),
```
ScottEilerman, line 852: I guess you could go down to `int8` or `bool` to save some memory, although I don't know if there's a compatibility issue with xesmf down the line and what dtypes it expects, and it probably doesn't matter much in the scheme of things.

roms_tools/setup/boundary_forcing.py
```diff
@@ -327,18 +485,36 @@ def __post_init__(self):
...
-                        processed_fields[var_name] = lateral_regrid.apply(
+                        processed_fields[var_name] = lateral_regrid_vector.apply(
                             bdry_data.ds[var_names[var_name]["name"]]
                         )
 
                     if self.adjust_depth_for_sea_surface_height:
                         # Regrid sea surface height ('zeta') onto a 2-cell-wide margin.
                         # This is needed to correctly infer depth coordinates at u- and v-points along the boundary.
-                        zeta_vector = lateral_regrid.apply(
+                        # 'zeta' is a tracer, so it uses the tracer mask (not the
```
ScottEilerman: I got a little confused about the `tracer` terminology here (and above) -- shouldn't it be `scalar` (as an opposite of vector)? The usage of tracer in the ocean modeling context makes it extra confusing.

roms_tools/setup/boundary_forcing.py
```diff
@@ -251,15 +333,53 @@ def __post_init__(self):
...
+                "the source first (the regrid is plain bilinear).",
+                prefill,
+            )
+        elif user_set_extrap and not use_xesmf:
+            logging.info(
+                "extrap_method/extrap_kwargs are ignored because the scipy regrid "
+                "engine is in use (xESMF unavailable or regrid_method='scipy'); "
+                "the source is nearest-neighbor pre-filled instead."
+            )
+
```
ScottEilerman, line 340: I'm wondering if it's possible to isolate all of the extrap/prefill combination resolution and logging in a method to declutter the `__post_init__`. `self._resolve_prefill_options()` had the right idea but there's still a lot leftover.

roms_tools/setup/utils.py
```diff
@@ -38,6 +38,190 @@
...
+
+#: Per-method allowed keys for ``prefill_kwargs`` / ``extrap_kwargs``. A method
+#: not listed here (e.g. ``None``) accepts no kwargs.
+PREFILL_ALLOWED_KWARGS: dict[str, frozenset[str]] = {
+    "2d_lateral_fill": frozenset(),
+    "nearest_neighbor": frozenset(),
+    "nearest_s2d": frozenset(),
+    "inverse_dist": frozenset({"num_src_pnts", "dist_exponent"}),
+    "creep_fill": frozenset({"num_levels"}),
+}
```
ScottEilerman: For constraining/validating inputs, at a minimum, we should use one or more `StrEnum` to avoid repeating the same strings all throughout the codebase and to constrain "you must pick one of these valid options" for free. We could go a step or two further and build small dataclasses or pydantic models for the prefill/extrap methods that map the method names to actual methods, accepted kwargs, compatibility with xesmf, etc. I think it would reduce a lot of the confusing if/else and validation boilerplate, both in this module and throughout the classes, if we could define the methods and their properties in one place. I'll see if I can find a similar example to what I'm thinking, or sketch one up.

ScottEilerman (commented):
A few initial suggestions, though I want to find some examples for my last comment and try to think a bit more about how to consolidate things.

roms_tools/setup/utils.py
kthyng: This is covered by your new and upcoming work I think.

roms_tools/setup/boundary_forcing.py
kthyng: This will happen in your upcoming work I think

roms_tools/setup/boundary_forcing.py
kthyng: You're right that is weird, I updated this

roms_tools/datasets/lat_lon_datasets.py
kthyng: I left this alone but let me know if you want it to change

ScottEilerman (approved):

PR #618, opened 2026-06-16
SurfaceForcing is daskified!

blsaenz:
RE-arranges some computations, and has some re-writes (same outcome, different variable usage, xarray notation) of computations to prevent dask from loading data prematurely before the save step.

- [X] Passes `pre-commit run --all-files`
- [X] Changes are documented in `docs/releases.md`


blsaenz:
@kthyng adding you here so you can see the diffs that helped with daskifying in this class

kthyng:
@blsaenz Thanks, this looks helpful. Can you explain at a high level what changes were important? I see, for example, a lot of change in the way variables are being assigned and related tasks in `lat_lon_datasets.py` — are those important or incidental? What are the important things?

blsaenz:
@kthyng I think the changes can be boiled down to:

- Avoiding checks for nans, or operations that use np.any() since that forces a load.
- Using some form of .assign() on xarray datasets, or doing whatever it takes such that xarray does not think you are changing/accessing the underlying data.  The rules around this seem opaque to me, but I think the lesson is to be careful whenever you have scalars involved, and that +=, *= (and their long-form equivalents) are treated as data reassignment.
- Method re-use (in this case a flag for return_coords_only or something similar) can have non-lazy consequences. In choose_subdomain, we were operating on the entire dataset, then grabbing the coordinates, but we should have been down-selecting for coordinates to avoid loading the entire dataset.

kthyng:
@blsaenz Ok thanks. I think this will have consequences across the big classes.

ScottEilerman (approved):

PR #620, opened 2026-06-22
BGC density interpolation

kthyng:
Add density-space vertical interpolation for BGC tracers

BGC tracers follow isopycnals, so interpolate them onto the model's
vertical grid in density space rather than depth space (per UCLA's
Add_bgc2ini.m). New `use_density_interpolation` parameter (default True)
on InitialConditions and BoundaryForcing.

- Source density is built from the BGC dataset's own temperature/salinity
  (a single source pair, declared in opt_var_names under the stable keys
  temp_bgc/salt_bgc; the unified dataset maps these to temp_WOA/salt_WOA).
- Target density uses the model's T/S: physics in the same object for
  InitialConditions; a physics BoundaryForcing passed as physics_forcing=
  for BoundaryForcing.
- When a BGC source lacks T/S (e.g. CESM), interpolation logs and falls
  back to depth space.
- Adds gsw dependency (sigma-0 via TEOS-10).

Task: https://cworthy.atlassian.net/browse/CSD-869

- [x] Tests added


kthyng:
The default is to keep depth-interpolation, but I have started density-based interpolation comparisons. It can look a bit strange. One example of the boundary from the end to end notebook both ways (orange lines show locations of detailed comparisons that are in my notes):

Depth-interpolation:
image

Density-interpolation:
image

blsaenz (approved):
Looks great, thank you

blsaenz:
I didn't fully absorb pytests - I am assuming the old tests passed, and new ones were added, rather than needed to regenerate data for the old tests since this pathway is optional?

kthyng:
@blsaenz 

>  I am assuming the old tests passed, and new ones were added, rather than needed to regenerate data for the old tests since this pathway is optional?

Essentially yes: two new .zarr were added to support tests using the density interpolation, and the two old .zarr (bgc_boundary_forcing_from_unified_climatology.zarr and initial_conditions_with_unified_bgc_from_climatology.zarr) were regenerated because I updated the test file coarsened_UNIFIED_bgc_dataset.nc.

kthyng:
I think I forgot to reference the original PR here: https://github.com/CWorthy-ocean/roms-tools/pull/607

PR #623, opened 2026-06-23
Update scipy pin and test data, and drop 3.11 support

ScottEilerman:
Our CI validation tests started failing last week for a particular validation test even though we hadn't changed anything with that data or methodology. I traced it back to scipy's 1.18 release on June 19 -- it seems to have very slightly changed results (but only in one test). This PR pins to the new scipy version and updates the validation data. If this causes problems in other parts of the stack, we could alternatively revert this commit and pin to scipy < 1.18. We also need to drop python 3.11 support to make this update, as scipy 1.18 only supports >= 3.12.

Contributors will need to update their development environment for tests to pass locally. Little-to-no disruption is expected for users.

- [ ] Closes #xxxx
- [ ] Tests added


blsaenz (approved):
Looks good.  Do we want to expand support for 3.14+ ?

ScottEilerman:
> Looks good. Do we want to expand support for 3.14+ ?

I can add it to the test matrix and see if it passes!

ScottEilerman:
Looks like there was a conflict with 3.14 and our numpy pin. Going to leave 3.14 out for now.

PR #624, opened 2026-06-24
Update releases.md

ubbu36:
adding links for Ulla's PRs in releases.md


smaticka:
Your precommit test was failing so I just pushed the change. Looks good now @ubbu36 , thanks for fixing! I think all of the other failing tests are because the UBGC file in roms-tools-test-data was updated before the ROMS-Tools was pushed.

smaticka (approved):

PR #625, opened 2026-06-26
Glofas integration

abigale-wyatt:
## GloFAS River Dataset Integration

This PR adds support for the GloFAS v4.0 global river discharge dataset as an alternative to Dai & Trenberth, along with performance improvements and bug fixes to the river forcing pipeline.

## What's new

- `GloFASRiverDataset` and `DaiRiverDataset` classes with dataset-appropriate coastal snap buffer defaults (50 km and 200 km respectively). `source={"name": "GLOFAS", "path": "..."}` is now a valid `RiverForcing` input alongside `"DAI"`.
- `coast_snap_buffer_km` and `domain_edge_buffer` fields on `RiverForcing`
- `latlon_to_xyz`, `build_kdtree_from_latlon`, `query_kdtree_nearest` utilities in `setup/utils.py`
- Updated `docs/river_forcing.ipynb` with a GloFAS usage example
- Improved gap-year handling in dynamic BGC calendar-year interpolation: partial boundary years (< 25% of a year's time steps) are now treated as missing and filled from the nearest full year instead of producing unrealistic annual means.

## Performance

- Replaced O(n×m) great-circle distance loops with cKDTree queries throughout 
- Added result caching in `Rivr2oRiverBGCDataset`
- Replaced the per-tracer assignment loop in `_apply_bgc_tracers` with `xr.concat` (~10x speedup), and added a progress bar during `.compute()` 
- `RiverDataset.load_data()` now slices the time axis with a 1-day buffer before triggering `.compute()`, avoiding loading full multi-year files into memory.

## Bug fixes

**Sorting and overlap handling (introduced to support GloFAS)**
- Rivers are now re-sorted by mean discharge after overlap handling, since absorbed rivers are zeroed out during overlap creation making the pre-overlap sort order meaningless
- `nriver` coordinates are reassigned as sequential 1-based integers after the final sort, ensuring YAML round-trips produce identical output regardless of discovery path
- Fixed `ValueError` when reusing `indices=` from a prior run — synthetic `overlap_*` rivers saved in a YAML file are now stripped before `extract_named_rivers` and regenerated by `_handle_overlapping_rivers`
- Overlap river names are now derived from `sorted(river_list)[0]` for deterministic naming regardless of dict iteration order
- Added early return in `sort_by_river_volume` when station list is empty
- Fixed river name/index misalignment after sorting in `extract_relevant_rivers`

**xarray compatibility**
- Fixed `where(drop=True)` in `extract_named_rivers` collapsing unrelated dimensions in some xarray versions — replaced with `isel` + boolean mask
- Fixed `river_tracer` dimension ordering to `(ntracers, river_time, nriver)` to match ROMS expectations
- Fixed `river_tracer` dtype — `_apply_bgc_tracers` now casts to `float32` and uses positional `isel` to avoid float64 overflow and coordinate lookup failures

**Plotting**
- Fixed `plot_locations()` showing identical points in both panels 
- Fixed `plot()` shape mismatch for tracer variables

## Tests added
- 12 new tests covering GloFAS source validation, dataset-layer
  extraction (bounding-box filter, snap-buffer threshold, empty-domain
  error), nearest-coastal-cell assignment, an end-to-end synthetic-file
  load, and a daily/multi-year climatology case; GloFAS fixture added to the
  shared parametrized test suites Dai fixtures already ran under.

## Review Checklist:

- [x] Tests added


conftest.py
```diff
@@ -26,6 +26,7 @@
     UnifiedBGCSurfaceDataset,
 )
 from roms_tools.setup.nesting import align_grids, make_nesting_info
+from roms_tools.tests.glofas_test_utils import write_glofas_file
```
ubbu36: this import fails for me, and hence the entire pytest suite fails. There is no 'roms_tools/tests/glofas_test_utils.py' in the repo

roms_tools/setup/river_forcing.py
```diff
@@ -713,11 +808,17 @@ def _handle_overlapping_rivers(self, ds: xr.Dataset) -> xr.Dataset:
             combined_river_volumes = []
             combined_river_tracers = []
 
+            name_to_idx = {name: i for i, name in enumerate(ds.river_name.values)}
             for i, (idx_pair, river_list) in enumerate(overlapping_rivers.items()):
+                name = "overlap_" + sorted(river_list)[0].replace("GloFAS_", "")
```
ubbu36: This seems to be a GLOFAS specific overlap handling in river_forcing, which should be data-agnostic if possible.

roms_tools/setup/river_forcing.py
```diff
@@ -256,11 +269,18 @@ def __post_init__(self):
...
             target_coords = get_target_coords(self.grid)
+            target_coords["mask"] = self.grid.ds.mask_rho
             # maximum dx in grid
             dx = (
                 np.sqrt((1 / self.grid.ds.pm) ** 2 + (1 / self.grid.ds.pn) ** 2) / 2
             ).max()
-            original_indices = data.extract_relevant_rivers(target_coords, dx)
+
+            kwargs = {"domain_edge_buffer": self.domain_edge_buffer}
+            if self.coast_snap_buffer_km is not None:
```
ubbu36, line 291: None on RiverForcing cannot mean “disable the coastal filter” because that would require passing coast_snap_buffer_km=None into extract_relevant_rivers. Also, the docstring on coast_snap_buffer_km says Dai defaults to None; it actually defaults to 200 km on DaiRiverDataset. Consider clarifying the docstring and/or adding a brief comment on why domain_edge_buffer is always passed but the snap buffer is not.

roms_tools/setup/river_forcing.py
ubbu36, line 1315: This plotting functionality looks wrong with GLOFAS when I test it in a notebook. According to Claude that is because: "left panel title says “Original river locations”, but it plots lon_rho/lat_rho at original_indices, which come from extract_relevant_rivers step 3 (nearest grid cell, including ocean). It does not plot the source mouth coordinates (lat_mou/lon_mou from Dai/GloFAS)." 

Either way, we should fix this plotting functionality to be consistent with GLOFAS

roms_tools/setup/utils.py
```diff
@@ -256,6 +255,10 @@ def interpolate_dynamic_bgc_by_calendar_year(
     years are linearly interpolated along the year axis, and the result is
     broadcast back to every ``river_time`` step. Leading and trailing NaN years
     are left unchanged.
+
+    # TODO: years with very few time steps (e.g. a 1-day buffer before start_time of Jan 1)
```
ubbu36: This is actually crucial to fix because we have this padding functionality, so the first datapoint looks super weird in any data series that has time-varying tracers (i.e. rivr2o tracers). We could just have the padding time-step be a copy of the first time step in the actual simulation window? In either case, how it is now is not acceptable in the case of glofas+rivr2o

roms_tools/tests/test_setup/test_river_forcing.py
```diff
@@ -901,3 +948,45 @@ def test_roundtrip_yaml(self, river_forcing_with_rivr2o_bgc, tmp_path):
...
+        filepath = tmp_path / "test_yaml_glofas"
+        river_forcing_with_glofas.to_yaml(filepath)
+
+        caplog.clear()
+        with caplog.at_level(logging.INFO):
+            river_forcing_from_file = RiverForcing.from_yaml(filepath)
+
+        assert "Use provided river indices." in caplog.text
+        assert river_forcing_with_glofas == river_forcing_from_file
+        filepath.unlink()
```
ubbu36, line 1001: Several important paths are still untested. Suggest adding the following tests:

1. Source validation

GLOFAS without path should raise — _normalized_source() requires a path for all sources except Dai, but there is no test analogous to test_bgc_rivr2o_requires_path for GloFAS.

2. extract_relevant_rivers (dataset layer)

Suggest unit tests for:

Bounding-box pre-filter (stations inside/outside domain)
50 km coastal snap buffer default for GloFASRiverDataset (stations just inside vs outside buffer)
cKDTree nearest-cell assignment
Empty-result behavior (no stations in domain → {} and empty dataset)

3. End-to-end GloFASRiverDataset load

TestGloFASRiverDataset only checks defaults/metadata — no test runs GloFASRiverDataset.__post_init__ on a synthetic GloFAS-format NetCDF (via write_glofas_file once that helper is committed).
Should assert variable mapping (FLOW, riv_name, vol_stn, CF time), time subsetting, and deduplication of duplicate station names.

4. Time / climatology behavior

Dai climatology is covered by existing fixtures (river_forcing, river_forcing_no_climatology, zarr regression).
No GloFAS-specific tests for daily time series, multi-year windows, or convert_to_climatology (never / always / if_any_missing) with daily GloFAS data.
CI should pin at least one daily multi-year case.

5. Integration fixture coverage

river_forcing_with_glofas is only used in TestRiverForcingWithGloFAS and ROUNDTRIP_FIXTURES.
It is not in shared parametrized suites such as STANDARD_RIVER_FIXTURES or INITIALIZATION_FIXTURES, so GloFAS does not get the same initialization, tracer, coast-location, or hash-regression coverage as default Dai fixtures.
Consider adding it to those lists and/or to test_validation.py

ubbu36 (changes requested):
Looks good! The most important comment I left is regarding failure of the pytest suite because it tries to import a module that doesn't exist. Then we need to figure out what to do with the padding when using glofas+rivr2o, and update the plotting functionality to be consistent with GLOFAS, and I added some suggestions for most tests to be consistent with Dai and rivr2o tests.

Beyond that, we also need to address the following:

docs/methods.md still describes only Dai.
docs/features.rst lists Dai but not GloFAS.
docs/api.rst has no GloFAS mention.

Let me know if you have any questions to any of these things.

roms_tools/setup/river_forcing.py
abigale-wyatt: I've removed the left panel entirely. It's not very useful for GloFAS (which is already coastally located) and I think it makes more sense to plot locations once anyway. I've also removed the computation for calculating "original_indices" since its no longer needed. 

For context, the issue is that the "original indices" were an attribute on the dataset and not processed similarly to the data itself. Therefore in a large dataset like GloFAS it didn't look much like the river forcing file after overlapping, filtering and sorting making the plot misleading and not very useful.

roms_tools/setup/river_forcing.py
abigale-wyatt: Fixed the docstrings. `None` now clearly means 'use the dataset's default', not 'disable the filter'... disabling is not possible

roms_tools/setup/river_forcing.py
abigale-wyatt: Ah whoops. Fixed by adding `_river_name_prefix`

conftest.py
abigale-wyatt: added `roms_tools/tests/glofas_test_utils.py` to the commit. You should be able to import it now.

abigale-wyatt:
> Beyond that, we also need to address the following:
> 
> docs/methods.md still describes only Dai. docs/features.rst lists Dai but not GloFAS. docs/api.rst has no GloFAS mention.

Do we have a list somewhere of which docs I should have checked? `docs/releases.md` was specifically listed on the checklist so I updated that but didn't realize I needed to check these other files.

blsaenz:
> > Beyond that, we also need to address the following:
> > docs/methods.md still describes only Dai. docs/features.rst lists Dai but not GloFAS. docs/api.rst has no GloFAS mention.
> 
> Do we have a list somewhere of which docs I should have checked? `docs/releases.md` was specifically listed on the checklist so I updated that but didn't realize I needed to check these other files.

I think there was a checkbox (from Nora) that listed those things ... we may have deleted the features.rst because we weren't really adding new 'features,' just options on current features.  We already looked at api.rst and decided that it will auto-generate from and new method's docstrings (we don't have any new files or directories for the rst process to scan ducting documentation generation), and erased that checkbox and not applicable.  So, I think methods.md is the main thing to check.

roms_tools/setup/utils.py
abigale-wyatt: alright, I thought maybe this was outside the scope of this PR, but I went ahead and fixed it in `interpolate_dynamic_bgc_by_calendar_year`. Now, any years that have less than a quarter of the year will be filled from the adjacent year rather than using a flow rate that could be highly seasonal to calculate the concentration.

roms_tools/datasets/river_datasets.py
```diff
@@ -525,66 +542,135 @@ def sort_by_river_volume(self, ds: xr.Dataset) -> xr.Dataset:
 
         return ds
 
-    def extract_relevant_rivers(self, target_coords, dx):
-        """Extracts a subset of the dataset based on the proximity of river mouths to
-        target coordinates.
+    def extract_relevant_rivers(
```
blsaenz, line 548: It looks like below we are overriding methods jus tto change the argument defaults, then calling super() on the class method.  This creates a lot of lines of code and is pretty are to read to undetstand what is happending.  Here is what I suggest: On the super class (here RiverDataset) setup the class with constants for the defautl values, and have this method use the constants internally as defaults.

Then, the subclasses only need to have difference constants, rather than a whole block of code with an override that sometimes obsuscures the return value(s) etc.  See further comments below for where to replace those overrdies.  See:    

```
class RiverDataset:
    # 1. Define global defaults on the base class
    COAST_SNAP_BUFFER_KM: float = None
    DOMAIN_EDGE_BUFFER: int = 20

    def extract_relevant_rivers(
        self,
        target_coords: dict,
        dx: float,
        *,
        coast_snap_buffer_km: float | None = None,
        domain_edge_buffer: int | None = None,
    ) -> None:
        # 2. Fall back to instance/class attributes if not explicitly passed
        coast_buffer = (
            coast_snap_buffer_km 
            if coast_snap_buffer_km is not None 
            else self.COAST_SNAP_BUFFER_KM
        )
        edge_buffer = (
            domain_edge_buffer 
            if domain_edge_buffer is not None 
            else self.DOMAIN_EDGE_BUFFER
        )
```

roms_tools/datasets/river_datasets.py
```diff
@@ -634,9 +722,77 @@ def extract_named_rivers(self, indices):
...
+            "ratio": "ratio_m2s",
+            "name": "riv_name",
+        }
+    )
+    opt_var_names: dict = field(default_factory=lambda: {"vol": "vol_stn"})
+    name_prefix: str = field(default="GloFAS_", init=False)
+    """Prefix prepended to GloFAS river names, stripped back off when naming
+    synthetic ``overlap_`` rivers."""
+
+    def extract_relevant_rivers(
```
blsaenz: If the base class is updated as above, then this def override can be completely replaced with this:

```
    COAST_SNAP_BUFFER_KM: float = 50.0
    DOMAIN_EDGE_BUFFER: int = 20
```

roms_tools/datasets/river_datasets.py
```diff
@@ -699,6 +868,26 @@ def add_time_info(self, ds: xr.Dataset) -> xr.Dataset:
 
         return ds
 
+    def extract_relevant_rivers(
```
blsaenz: If the base class is updated as above, then this def override can be completely replaced with this:

```
    COAST_SNAP_BUFFER_KM: float = 200.0
    DOMAIN_EDGE_BUFFER: int = 20
```

roms_tools/datasets/river_datasets.py
```diff
@@ -525,66 +542,135 @@ def sort_by_river_volume(self, ds: xr.Dataset) -> xr.Dataset:
...
-                self.dim_names["station"], "eta_rho", "xi_rho"
+        station_dim = self.dim_names["station"]
+
+        # Build candidate point set — coastal cells only if mask provided, else all grid cells
+        target_lon_np = target_coords["lon"].values
+        target_lat_np = target_coords["lat"].values
+        eta_rho, xi_rho = target_lon_np.shape
+
+        if target_coords.get("mask") is not None:
+            mask_rho = target_coords["mask"].values
```
blsaenz, line 630: This section of code is doing a lot, without comments.  It took a while to figure out what you were trying to do, sum.  I think either put in some code comments about the process of making tree_lat,tree_lon, or use an available method which has it's own documentation and describes what is happening.  I'm pretty sure faces is just a convolution (see below)?  Also, look at DataArray.shift() - you could keep this all in xarray with that maybe.

```
from scipy.ndimage import convolve

# 1. Define the 4-neighbor kernel (North, South, East, West)
kernel = np.array([[0, 1, 0], 
                   [1, 0, 1], 
                   [0, 1, 0]])

# 2. Extract mask and convolve
mask_rho = target_coords["mask"].values
faces = convolve(mask_rho, kernel, mode="constant", cval=0.0)
```

roms_tools/setup/utils.py
```diff
@@ -278,9 +281,19 @@ def interpolate_dynamic_bgc_by_calendar_year(
             )
 
         tagged = values.assign_coords(calendar_year=calendar_years)
-        annual = tagged.groupby("calendar_year").mean(dim=time_dim, skipna=True)
-        annual = annual.reindex(calendar_year=full_years)
-        annual = annual.interpolate_na(dim="calendar_year", method="linear")
+        # Treat years with less than a quarter of the year as missing and interpolate.
```
blsaenz, line 340: Is this in the docs?  I still need to look at notebooks/docs...

blsaenz (approved):
Nice - just a few code changes from me at this point. I did not have a change to review the notebooks, but you and Ulla spent time on those so.  I'll look later, but I think with these changes (and fixing any tests) we can merge.

ubbu36 (approved):

blsaenz (changes requested):
I accidentally made two reviews, please look at the comments from the last one!

roms_tools/datasets/river_datasets.py
abigale-wyatt: Yes. That makes much more sense. Done.

roms_tools/datasets/river_datasets.py
abigale-wyatt: done

roms_tools/datasets/river_datasets.py
abigale-wyatt: done

roms_tools/datasets/river_datasets.py
abigale-wyatt: I moved the convolution to a utils function `find_coastal_cells`, that way I could add more comments in a doc string that carry with the function. I also tried to write clearer comments in the `extract_relevant_rivers` function.

roms_tools/setup/utils.py
abigale-wyatt: It could be documented better. You can see @ubbu36 's review earlier where I added that in so that we didn't get strange initial or final values due to partial years. When we do a docs overhaul, this should be addressed more clearly.

abigale-wyatt:
All comments have been implemented.

PR #626, opened 2026-06-26
Add mixed layer dep-anchored BGC vertical interpolation (density_mld)

kthyng:
Adds a third BGC tracer vertical-interpolation method, density_mld, alongside the existing depth and density options on both InitialConditions and BoundaryForcing.

density_mld identifies the mixed layer depth (MLD) in both the source and target density fields and interpolates the tracer linearly in depth within two MLD-anchored segments: the source mixed layer is scaled so its MLD matches the target's, and below the MLD the tracer is interpolated 1:1 in depth. This fixes the surface artifacts that density (isopycnal) interpolation produces in the mixed layer — where density is nearly uniform, isopycnal mapping is ill-posed and yields discontinuities/NaNs — while preserving mixed-layer structure across the regrid.

API change (with deprecation)
* New string selector bgc_interpolation_method ("depth" | "density" | "density_mld"), default "depth" — no behavior change for existing configs.
* use_density_interpolation (bool) is deprecated. A shim maps True → "density", False → "depth" with a DeprecationWarning; old YAML still deserializes. New YAML emits only bgc_interpolation_method.

compute_mld is generalized for postprocessing use: it now uses an xgcm.transform engine (consistent with xroms), is orientation-agnostic, and takes a reference_depth parameter (default 10 m, de Boyer Montégut 2004; pass 0 for the surface-referenced xroms convention). Documented with references + the non-monotonic-profile caveat, and added to docs/api.rst (with compute_potential_density).

Below the MLD the map is 1:1 in depth (not bottom-anchored), so a shallow target against a deep source draws from comparable absolute depths rather than compressing the full source column into a thin shelf.
Columns lacking a resolved mixed layer — fully-mixed source or target, NaN/degenerate MLD — fall back to the identity map (= plain depth interpolation) per column, so stratified and degenerate columns coexist in one vectorized call. Example added to Reading Roms Ouput notebook.

Testing
* New unit + integration tests in test_utils.py, test_initial_conditions.py, test_boundary_forcing.py: exact-value checks of the warp, per-column mixed/stratified behavior, fully-mixed-source and fully-mixed-target degeneracy to depth, deprecation-warning and invalid-method paths, compute_mld known-crossing / orientation / reference-depth cases.
* Full setup suite passes locally (the one pre-existing test_check_results[initial_conditions_with_bgc_from_climatology] failure is unrelated CESM-fixture drift on main).
* docs/initial_conditions.ipynb and docs/boundary_forcing.ipynb updated to the new API and re-executed.

- [ ] Tests added


kthyng:
Example interpolation comparisons:

Alkalinity, one boundary:

image

Diff between depth-interpolation and MLD-respecting-interpolation:
image



Alkalinity, Another boundary:
image


Nitrate, one example:
image

Diff between depth-interpolation and MLD-respecting-interpolation:
image

roms_tools/setup/initial_conditions.py
```diff
@@ -113,17 +115,30 @@ class InitialConditions:
...
+          features, and avoids the surface degeneracy of pure density space.
+
+        ``"density"`` and ``"density_mld"`` only apply when ``bgc_source`` is provided,
+        the physics source is a lat/lon dataset (not a ROMS restart), and the BGC source
+        carries temperature/salinity (e.g. the unified dataset's ``temp_WOA``/
+        ``salt_WOA``); otherwise interpolation falls back to depth space and notes in
+        the log. Interpolation uses ``xgcm.Grid.transform`` with the linear method
+        inside the source range and edge-value extrapolation outside
+        (``mask_edges=False``).
     use_density_interpolation : bool, optional
```
blsaenz: I recommend we just get rid of use_density_interpolation, and machinery/comments around deprecation.  We have only had it around for a few days, and I don't think any tooling is very dependent on it (Scott's Forge refactor can deal with this change I'm sure).  ?

roms_tools/setup/utils.py
```diff
@@ -727,6 +729,265 @@ def _compute_density_coord(
...
+BGC_INTERPOLATION_METHODS = ("depth", "density", "density_mld")
+
+# Mixed-layer-depth detection defaults (density-threshold criterion, de Boyer
+# Montégut et al. 2004): the MLD is the depth at which potential density first exceeds
+# the value at ``MLD_REFERENCE_DEPTH`` by ``MLD_DENSITY_THRESHOLD``.
+MLD_DENSITY_THRESHOLD = 0.03  # kg/m^3
+MLD_REFERENCE_DEPTH = 10.0  # m
+
+
+def compute_mld(
```
blsaenz, line 748: How is the speed on this for a big/long domain?  I suppose if it is fully dask lazy and resolves to a numpy-only graph it could be reasonably performant.

roms_tools/setup/initial_conditions.py
kthyng: Yes I would kind of like to get rid of it especially if we don't plan to use it but I stopped because I thought, well, maybe someone would want to use it in the future? Do you think that could be the case?

blsaenz: I meant just remove the extra code around deprecation of the "use_density_interpolation" kwarg (but keep the actual functionality with the new kwarg input).  I feel like there's no use handling the deprecation with a bunch of lines of translation code and a bunch of extra notes/comments when we put in the feature in just a week ago.

kthyng: oh ok sure

kthyng: all references and code for use_density_interpolation are removed now.

roms_tools/setup/utils.py
kthyng: I am going to come back to timing it due to HPC issues at the moment. The default remains depth so this should not impact performance for now.

PR #629, opened 2026-07-02
Surface forcing download

kthyng:
## Summary

Three related improvements to `SurfaceForcing`, plus ERA5 download docs:

### 1. Time-chunked climatology interpolation (memory)
The interpolation of a climatology onto the forcing time axis now proceeds in blocks instead of interpolating the whole target axis in one call — so it no longer pulls the full time range into memory as a single `(time, y, x)` array. This fixes two things:
- **Radiation-correction memory:** `correct_radiation=True` on long/multi-year records previously materialized the entire corrected field at once (could OOM on large grids); it now streams in bounded chunks.
- **Validation memory:** per-timestep access (e.g. the `isel(time=0)` used in validation) only computes the block it falls in, so validation is cheap and no longer needs to be bypassed.

### 2. `prefill` + `regrid_method` options from `BoundaryForcing` (speed)
Brings the `BoundaryForcing` regrid/prefill machinery (#616) into `SurfaceForcing` for all types (`physics`/`bgc`/`restoring`): `regrid_method` (`"auto"`/`"xesmf"`/`"scipy"`), `prefill` (+ `prefill_kwargs`), and `extrap_method` (+ `extrap_kwargs`).
- **Default change for speed:** the horizontal regrid now uses masked xESMF bilinear + inverse-distance extrapolation (scipy + nearest-neighbor pre-fill fallback without xESMF), replacing the AMG Poisson 2D lateral fill + scipy interpolation — the old fill was memory-intensive and slow, especially on large multi-year domains. Output shifts slightly on defaults; `prefill="2d_lateral_fill", regrid_method="scipy"` reproduces the legacy output.
- The xESMF source grid is built on the source's own lat/lon dim names so xESMF matches input axes by name (order-independent), which is what lets every bgc/restoring climatology source regrid cleanly.
- On the default (extrapolating) path, a grid that extends beyond the source coverage now raises a clear coverage error instead of silently extrapolating.
- The choices are recorded in dataset metadata and round-trip through `to_yaml`/`from_yaml`.

### 3. ERA5 download documentation
New "Downloading ERA5 data" section in the datasets notebook showing how to pre-save ERA5 surface forcing (one file per day, NetCDF or Zarr) from the analysis-ready ARCO cloud store for offline/local use.

### Tests & docs
Physics regression fixtures regenerated for the new default; bgc fixtures pinned to `scipy` + `2d_lateral_fill` for byte-reproducibility (matching `BoundaryForcing`); added default-path tests for physics and bgc/restoring and a coverage-error test. Surface-forcing notebook and release notes updated.

### Breaking change
Surface forcing output shifts slightly for users on the defaults due to the change from AMG Poisson 2D lateral fill + scipy to masked xESMF bilinear + extrapolation. Pass `prefill="2d_lateral_fill", regrid_method="scipy"` to reproduce the legacy output.

## Timing Tests

On 0.5 an Anvil node this new code took to run 1 year of forcing files on Abigale's large Pacific domain:
* streaming data: 8 min/month
* previously-saved netcdf files fronted with a kerchunk file: 7 min/month

Before these changes it died during the object setup. If I used `bypass_validation=True` and `correct_radiation=False` then it would make the object fine and take 12 min/month. I haven't seen a big enough difference in time between streaming and pre-saving the files (whether netcdf+kerchunk, or direct zarr) to justify the disk space for pre-saving.

- [x] Tests added


ScottEilerman (approved):
Looks good!

PR #630, opened 2026-07-02
Add option to download end-to-end example as Python script

NoraLoose:
In our [JOSS review](https://github.com/openjournals/joss-reviews/issues/10234#issuecomment-4856426979), we got the following comment:

> having the .ipynb download option is great, but since I am not much of a user of Jupyter notebook, I ended up copy/pasting the tutorial's cells into a Python script one section after another. Would it be possible to have a Python script download option ? Maybe not for all the tutorials but for the end-to-end workflow as a starting point.

This PR addresses this suggestion.


kthyng (changes requested):
I tried this out and it looks like a good setup, but I did have one issue. I ran the docs locally and when I clicked on the button for the script and then tried to run the script, I hit that one of the markdown lines wasn't appropriately commented:
"Let's first look at the snapshots of the physical variables."
Any idea why that line wasn't coded correctly?

kthyng:
That merge blocking language and red icon looks pretty serious. I don't mean it as harshly as that looks, geez!

NoraLoose:
Nice catch, @kthyng! Thanks for actually testing that python script (which I had not done, my bad!). The problem was that one cell was marked as "raw" rather than "markdown" in the original notebook. I just fixed that, so now it should work.

kthyng (approved):

kthyng:
Sorry for the delay @NoraLoose and the holdup! I went out of town after doing that. Thanks for the fix.

Issue #584, opened 2026-04-14
Optimization ideas for LatLonDataset longitude wrapping


ScottEilerman:
Currently, for source datasets that have global coverage, we are concatenating all variables with 2 additional copies of data in order to be able to easily interpolate across the longitude discontinuity (see [_concatenate_longitudes](https://github.com/CWorthy-ocean/roms-tools/blob/main/roms_tools/datasets/lat_lon_datasets.py#L2290C1-L2360C21)). Unless xarray is being very smart/lazy in the background, this seems like an inefficient way to handle this.

Some options we could consider, from minor to major overhauls:
* Utilize the grid's `straddle` param and only do longitude concatenation if needed
* Utilize the grids `straddle` param to create alternate longitude coordinates for both the grid and the LatLonDataset that are offset by 180 degrees (or other custom adjustment to fit the grid in a way that it doesn't straddle). Do any longitude-discontinuity-sensitive operations with the alternate longitude coordinates
* Utilize [xarray.Dataset.pad](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.pad.html) with `mode="wrap"` -- it may be more efficient than all of these manual concat operations
* Using pad or the current code, consider if a smaller wrap is sufficient (e.g. a linear interpolation should only need one data point wrapped on either side, not the whole grid -- out interpolation/regridding operations are more complex though)

NoraLoose:
> Currently, for source datasets that have global coverage, we are concatenating all variables with 2 additional copies of data in order to be able to easily interpolate across the longitude discontinuity

It's not quite as bad (but not optimal -- agreed). We don't always concatenate 2 additional copies. [This code block](https://github.com/CWorthy-ocean/roms-tools/blob/0419aac794efb312cd6d95fac9ee97e4c98f4222/roms_tools/datasets/lat_lon_datasets.py#L2424-L2443) tries to avoid concatenation if it's not necessary. But there is definitely room for improvement!

PR #631, opened 2026-07-07
Replace manual variable loop in _concatenate_longitudes with xr.Dataset.pad

steps-re:
## Summary

Closes part of #584.

The `_concatenate_longitudes` function previously extended the longitude dimension by iterating over every dataset variable and calling `xr.concat([field] * n_copies, dim=lon_name)` — one `concat` call per variable. This PR replaces that with a single `xr.Dataset.pad({lon_name: pad_width}, mode="wrap")` call, then immediately reassigns the correct shifted longitude coordinates (same logic as before).

### What changes

- The per-variable loop and `xr.concat`-of-copies is replaced by `Dataset.pad(mode="wrap")` — xarray's canonical API for periodic-boundary padding.
- The longitude coordinate is still reconstructed identically (`lon ± 360` via `xr.concat`), so output values are byte-for-byte equivalent to before.
- Public signature and semantics of `_concatenate_longitudes` are unchanged. All callers (`choose_subdomain`, `LatLonDataset.concatenate_longitudes`, `ERA5CorrectionDataset.match_subdomain`) work without modification.

### Why this helps

- One dataset-level `pad` call is cleaner and shorter than a per-variable dispatch loop.
- `Dataset.pad` preserves dask lazy evaluation by construction; the old `xr.concat` approach also worked lazily, but `pad` is the intended xarray primitive for this pattern.
- Sets up a natural path for a follow-on optimization: compute the minimum number of wrap points needed (rather than a full-width copy) and pass that to `pad`. That would be the memory win the issue is after.

### Testing

Ran all existing tests covering `_concatenate_longitudes` and `choose_subdomain`, in both dask and non-dask modes — all 10 pass:

```
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_concatenate_longitudes[lower-expected_lons0] PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_concatenate_longitudes[upper-expected_lons1] PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_concatenate_longitudes[both-expected_lons2] PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_invalid_end_raises PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_choose_subdomain_basic PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_choose_subdomain_straddle PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_choose_subdomain_wraps_negative_lon PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_choose_subdomain_respects_buffer PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_choose_subdomain_raises_on_empty_lon PASSED
roms_tools/tests/test_datasets/test_lat_lon_datasets.py::test_choose_subdomain_raises_on_empty_lat PASSED
```

Happy to benchmark or iterate — if you'd prefer this PR also tackle the minimum-pad-width optimization (so we stop copying the full global grid when only a small wrap is needed), I can add that.

Signed-off-by: Mike German <mike@stepsventures.com>

### DIFF
--- a/roms_tools/datasets/lat_lon_datasets.py
+++ b/roms_tools/datasets/lat_lon_datasets.py
@@ -2985,23 +2985,24 @@ def _concatenate_longitudes(
     use_dask: bool = False,
 ) -> xr.Dataset:
     """
-    Concatenate longitude dimension to handle global grids that cross
+    Extend the longitude dimension to handle global grids that cross
     the 0/360-degree or -180/180-degree boundary.
 
-    Extends the longitude dimension either lower, upper, or both sides
-    by +/- 360 degrees and duplicates the corresponding variables along
-    that dimension.
+    Wraps data from the opposite side of the globe using
+    :func:`xarray.Dataset.pad` with ``mode="wrap"``, then assigns the
+    correct shifted longitude coordinates.  Only the longitude dimension
+    is padded; all other dimensions and variables are left unchanged.
 
     Parameters
     ----------
     ds : xr.Dataset
-        Input xarray Dataset to be concatenated.
+        Input xarray Dataset to be extended.
     dim_names : Mapping[str, str]
         Dictionary or mapping containing dimension names. Must include "longitude".
     end : str
         Specifies which side(s) to extend:
-        - "lower": extend by subtracting 360 degrees.
-        - "upper": extend by adding 360 degrees.
+        - "lower": extend by subtracting 360 degrees (prepend wrapped data).
+        - "upper": extend by adding 360 degrees (append wrapped data).
         - "both": extend on both sides.
     use_dask : bool, default False
         Accepted for backward compatibility; no longer causes eager rechunking.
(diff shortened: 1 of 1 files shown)


steps-re:
Heads up on the 3.12 CI failure here: it was a flaky network timeout pulling the TPXO tidal fixtures from raw.githubusercontent.com (a `ReadTimeout` plus a truncated ~300 MB download / `IncompleteRead`), not the change itself. The identical 3.13 job passed on the same commit, 900 tests passed, and the `_concatenate_longitudes` and `choose_subdomain` tests all passed.

I pushed an empty commit to re-trigger CI, but the run is waiting on maintainer approval. Could you approve it when you get a chance? Happy to rebase if that's easier.

ScottEilerman (approved):
Looks like everything's green. Thank you @steps-re for the contribution!

PR #633, opened 2026-07-09
Remove CESM support in NB examples

smaticka:
Since we don't support acquisition of CESM data, we are removing it from the documentation as a means of a 'follow along example'. It still exists in the Boundary conditions notebook as a way to demonstration how to use the pre-fill on land masks for coarse source data. We might consider redoing this part when ROMS-Tools supports more BGC data sources. 


docs/boundary_forcing.ipynb
```diff
@@ -1459,16 +1459,15 @@
    "metadata": {},
```
kthyng, line 1459: <p>Looks like this needs to be reverted to CESM instead of glorys?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/CWorthy-ocean/roms-tools/pull/633/discussion/#comment-3554407495'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='72'/>

smaticka: <p>I intentionally changed this one to GLORYS. I didn't think it made sense to suggest to the user how to use script for CESM, when we aren't supporting it. I can specify "To use a pre-fill when building the <strong>physics</strong> boundary forcing.." to give it more clarity. </p>

kthyng: <p>Yep got it, good point.</p>

kthyng (approved):

PR #635, opened 2026-07-09
Bringing prefill and regrid options to Initial conditions class

kthyng:
BoundaryForcing ([#616](https://github.com/CWorthy-ocean/roms-tools/issues/616)/[#626](https://github.com/CWorthy-ocean/roms-tools/issues/626)) and SurfaceForcing ([#629](https://github.com/CWorthy-ocean/roms-tools/pull/629)) already select their horizontal regrid engine and optional source prefill through the shared processing_methods.RegridConfig, defaulting to a NaN-aware xESMF regrid. InitialConditions was the last full-domain consumer still hardwired to the legacy AMG Poisson fill + SciPy interpolation — so the interior initial state and the boundaries it must match at t=0 were built by different machinery. This PR closes that gap and factors the shared logic into one place.

Changes:
* InitialConditions gains prefill, prefill_kwargs, regrid_method, extrap_method, extrap_kwargs (same semantics as BoundaryForcing). No deprecated apply_2d_horizontal_fill alias — IC never exposed one.
* New shared helpers, so the mask-selection, prefill-branch, regridder-construction, and coverage-check logic live in exactly one place:
  * regrid.py: build_lateral_regridder, select_source_mask
  * setup/utils.py: check_source_coverage, apply_source_prefill, apply_scipy_fallback_fill
* BoundaryForcing and SurfaceForcing refactored onto these helpers — behavior-preserving. SurfaceForcing's private _check_source_coverage is removed in favor of the shared one.
* ROMS-restart guard: ROMSDataset has no prefill API, so for a "ROMS" source IC keeps its existing apply_lateral_fill path and the new options are ignored (with a logged note), mirroring the bgc_interpolation_method fallback convention.

Behavior change: 
The IC default is now prefill=None + regrid_method="auto" (masked xESMF bilinear + inverse-distance extrapolation; nearest-neighbor + SciPy fallback when xESMF is absent). Output shifts slightly for users on the defaults — pass prefill="2d_lateral_fill", regrid_method="scipy" to reproduce pre-v4 output byte-for-byte. On the default (extrapolating) path, a grid that extends beyond the source coverage now raises a clear error instead of silently extrapolating.

Testing: 
* New IC tests: default vs. legacy path, per-method prefill (inverse_dist/nearest_s2d/nearest_neighbor), source-coverage error, ROMS-source guard, and an explicit to_yaml→from_yaml round-trip of the new fields.
* BoundaryForcing and SurfaceForcing regression-hash fixtures are unchanged — the byte-identical guard confirming the shared-helper refactor is behavior-preserving.

Docs: 
* releases.md entry.
* initial_conditions.ipynb gains a short "Controlling the horizontal regridding" section documenting the options (referencing the Boundary Forcing tutorial for the full explanation).

Timing:
I was able to run a large domain (1.2 million horizontal cells and 100 vertical levels) on a full Anvil node in about 10 minutes simply inputting the GLORYS relevant file for the initial time and no chunks information so that the internal chunks management handled it. I tried a few versions of the unified BGC dataset with varied chunks and it did change the time but they are all fine. Running on half a node did not work — not enough memory I think.

- [x] Tests added


kthyng:
Also here is a comparison plot (original code - updated code) for the PACMED domain and temperature:
image

roms_tools/regrid.py
```diff
@@ -242,6 +242,77 @@ def apply(self, da, method="linear"):
...
+    -------
+    xarray.DataArray or None
+        The source mask, or ``None`` when masking does not apply.
+    """
+    if not (use_xesmf and prefill is None):
+        return None
+    tracer = ds["mask"] if "mask" in ds.data_vars else None
+    if is_vector:
+        return ds["mask_vel"] if "mask_vel" in ds.data_vars else tracer
+    return tracer
```
ScottEilerman: Looking for a sanity check again on "tracer" vs "scalar" as an opposite to "vector." I think I was convinced last time that in some case, the "tracer mask" made sense as a term, because it ended up only being used with actual tracers. But it still feels like the wrong generalized terminology for "the mask for all of the data on the regular grid, as opposed to the u/v offset grid."

ScottEilerman (approved):
I looked somewhat quickly given that the closer review was on the initial implementation, but it seems good in terms of extending it. I left a nit comment about the tracer/vector naming again, but feel free to ignore it if I'm not thinking about it correctly.

roms_tools/regrid.py
kthyng: ok I updated the tracer -> scalar language.

kthyng:
That failing test will be fixed by the xgcm pin in another branch.

PR #636, opened 2026-07-10
updates to TidalForcing using prefill and regridding

kthyng:
Extends the NaN-aware lateral regrid work already done for `BoundaryForcing` (#616), `SurfaceForcing` (#629), and `InitialConditions` (#635) to `TidalForcing`, the last masked-ocean-source class still on the legacy path.

**Stacks on #635** — branched from `initial_conditions_prefill_regrid` to reuse the shared regrid helpers it introduces.

### What changed

- `TidalForcing` gains `prefill`, `prefill_kwargs`, `regrid_method`, `extrap_method`, and `extrap_kwargs`, resolved once via `RegridConfig.from_options` (mirroring IC/BF/SF).
- **Default behavior flips**: masked xESMF bilinear regridding + inverse-distance destination extrapolation (nearest-neighbor prefill + scipy fallback without xESMF), replacing the always-on AMG Poisson 2D fill + scipy interpolation. Tidal output shifts slightly on defaults; `prefill="2d_lateral_fill", regrid_method="scipy"` reproduces the legacy output exactly.
- A grid that outruns the TPXO source coverage now raises a clear error on the default extrapolating path instead of silently extrapolating.
- The resolved `prefill`/`regrid_method`/`extrap_method` are recorded in `ds.attrs` and round-trip through `to_yaml`/`from_yaml`.
- `TidalForcing` now consumes the shared `build_lateral_regridder`/`select_source_mask`/`apply_source_prefill`/`apply_scipy_fallback_fill`/`check_source_coverage` helpers extracted in #635, rather than duplicating regrid logic. Since each TPXO sub-dataset (`h`/`sal`/`u`/`v`) carries its own location-correct mask, no tracer/velocity mask split is needed (simpler than BF/IC).

### Tests

- `conftest.py`'s `tidal_forcing` fixture is pinned to `prefill="2d_lateral_fill", regrid_method="scipy"` so existing regression reference data in `test_validation.py` stays valid unchanged.
- `test_unsuccessful_initialization_with_regional_data_due_to_nans` is likewise pinned to the legacy path, since the new default raises a coverage error instead of a NaN error for that grid/data combination.
- New tests: `test_tides_default_regrid_path`, `test_tides_legacy_path`, `test_tides_default_differs_from_legacy`, `test_tides_prefill_methods_produce_nan_free_output` (parametrized), `test_tides_source_coverage_error`, `test_tides_regrid_options_roundtrip_yaml`.

### Docs

- `docs/tides.ipynb` updated with an explanation of the new options.
- `docs/releases.md`: breaking-change entry (default flip) + new-feature entry (options) + note that `TidalForcing` now shares the extracted regrid helpers.

Example diff between new default and old:
image

- [x] Tests added


ScottEilerman (approved):
second verse, same as the first. thanks for these!

PR #639, opened 2026-07-16
add temp variability 

abigale-wyatt:
# Derive river temperature from ERA5 air temperature

## Summary

- Adds an optional `surface_forcing_source` parameter to `RiverForcing` — an ERA5 source dict (e.g. `{"name": "ERA5", "path": ...}`, the same shape as `SurfaceForcing`'s `source`). When set, river temperature is derived from ERA5 air temperature sampled at each river's mouth instead of the flat constant default (15.6°C).
- Air temperature is smoothed with a rolling-mean window (`river_temp_smoothing_window_days`, default 30 days) and floored at 0°C — a deliberately simple, uncalibrated placeholder for a river's thermal inertia, not a physically complete model.
- Sampling is nearest-neighbor via a KDTree, at the *mean* coordinate of each river's cells (one query per river, not once per grid cell), reading only the `Tair` variable narrowed to a small bounding box around the river mouths. This avoids building a full `SurfaceForcing` object (regrid, radiation/wind correction, other physics variables).
- Refactored ERA5 default-path/dataset-class resolution (`gs://`/`gcs://` → `ERA5ARCODataset`, else `ERA5Dataset`) out of `SurfaceForcing` into a shared `resolve_era5_source` helper in `lat_lon_datasets.py`, so `RiverForcing` reuses the same logic instead of duplicating it.

## Key design notes

- River temperature always samples **real**, non-climatological ERA5 data. Whether the *stored* result ends up climatological is governed by the discharge data (`convert_to_climatology`) and/or the BGC dataset's calendar requirements — not by `surface_forcing_source` itself. 
- A defensive coverage check (`_check_surface_forcing_source_coverage`) raises a clear error if the ERA5 source doesn't cover the requested time range, rather than silently producing bad output.

## Testing

  - Pure-function helpers: bounding-box construction, KDTree nearest-neighbor sampling, smoothing/flooring (including the two crash-fix regressions), and day-of-year climatology reduction with wraparound.
  - Source validation and resolution (`_normalized_surface_forcing_source`, `_resolve_surface_forcing_source`), including the ARCO-vs-local dataset-class dispatch and native longitude-convention conversion.
  - Integration coverage: default vs. ERA5-derived temperature, a multi-cell river's temperature verified to come from **one sample at the averaged coordinate** (confirmed with an exact analytically-predicted value, not just a finite/non-negative check), the climatology path, an insufficient-coverage error, and YAML round-tripping (verified with both dataclass `==` and `xr.testing.assert_equal` on the underlying dataset).
- Added regression tests for the shared `resolve_era5_source` helper and `SurfaceForcing`'s ERA5 default-path dedup.
- Full suite (`test_river_forcing.py`, `test_surface_forcing.py`, `test_lat_lon_datasets.py`): **257 passed, 4 skipped, 6 deselected**, no regressions.

## Docs

- `docs/releases.md` updated to describe the shipped `surface_forcing_source` API.
- `docs/river_forcing.ipynb` gains a demo section for the new parameter.

- [x] Tests added


roms_tools/setup/river_forcing.py
```diff
@@ -569,12 +845,255 @@ def _get_river_sample_coords(
...
+            ``ds`` with the ``temp`` tracer overwritten in place.
+        """
+        river_names = [str(name) for name in ds.river_name.values]
+        river_temp = self._get_river_surface_temp(
+            ds, river_names, smooth_window_days=self.river_temp_smoothing_window_days
+        )
+
+        logging.info("Computing river temperature from surface air temperature...")
+        with ProgressBar():
+            river_temp = river_temp.compute()
```
ubbu36, line 1106: this is the 2nd time you do .compute() with a progress bar - should only be done once, I think.

roms_tools/setup/river_forcing.py
```diff
@@ -569,12 +845,255 @@ def _get_river_sample_coords(
...
+
```
ubbu36, line 1107: Also, maybe it's appropriate to have a check for NaN's here? I *think* that without it, nans in era5 gets turned into 0s.

ubbu36 (commented):
Looks good overall. Apart from the two code changes I suggest. I also wonder about the name 'surface_forcing_source' - it's not very intuitive that it's to do with river temperature, specifically. I suggest renaming to river_temp_source. The fact that it originates from the surface forcing is more of a detail than the main point, which is to create river temperatures.

roms_tools/setup/river_forcing.py
abigale-wyatt: The other compute is for the volume discharge, this one is the only one for temperature and it's a step that takes a while so I left it in.

roms_tools/setup/river_forcing.py
abigale-wyatt: There shouldn't be NaNs in ERA5... but if someone had a partial file or something or if we use a different source eventually this is a good idea. I've added a check and a `ValueError`

blsaenz: NaN checks are very slow.  That is all.

abigale-wyatt:
> Looks good overall. Apart from the two code changes I suggest. I also wonder about the name 'surface_forcing_source' - it's not very intuitive that it's to do with river temperature, specifically. I suggest renaming to river_temp_source. The fact that it originates from the surface forcing is more of a detail than the main point, which is to create river temperatures.

So, I considered this. I could go either way. Initially, I used this naming because the intention was to match the surface forcing source specifically and that's a justification for using those air temps (which is a VERY rough way to get river temps). We could eventually use something like a full radiative balance from the surface forcing, or we could try some other methods of deriving river temps and then we definitely would want a rename. But yeah, maybe it makes sense to change it here anyway. @blsaenz thoughts?

roms_tools/setup/river_forcing.py
```diff
@@ -138,6 +143,197 @@ def _mask_invalid_dynamic_bgc_concentrations(
...
+    if tair.chunks is not None:
+        # Rolling reductions need the rolled dimension in a single chunk.
+        tair = tair.chunk({time_dim: -1})
+
+    smoothed = tair.rolling({time_dim: window_size}, center=True, min_periods=1).mean()
+
+    return smoothed.clip(min=0.0)
+
+
+def _bounding_box_with_buffer(
```
blsaenz, line 243: Don't change now, but this method it the kind of thing that makes code hard to read.  It is only two lines of code, and it is not re-used anywhere.  Making it a method then sort of cascades the need for pytest infrastructure to support, and need to read/grok 40 lines of code.  We can collapse this stuff in a revision hopefully down the road.

blsaenz (approved):

roms_tools/setup/river_forcing.py
abigale-wyatt: I removed that function and integrated it inline further down. Also removed the associated pytests.

PR #641, opened 2026-07-21
fixing dimensions and tests in river_forcing

ubbu36:
**Summary**
Fixes a regression in RiverForcing where river_tracer dimensions were written in the wrong order for ROMS NetCDF input.

ROMS expects (river_time, ntracers, nriver), but since the GloFAS integration (#625) the output used (ntracers, river_time, nriver).

**Changes**
Restore correct dimension order when allocating river_tracer in _create_river_forcing
Transpose after BGC tracer concatenation in `_apply_bgc_tracers`, since `xr.concat(..., dim="ntracers")` reintroduces the swapped order
Add regression test` test_river_tracer_dimension_order`
Regenerate expected zarr baselines for `river_forcing_with_bgc` and `river_forcing_no_climatology`


blsaenz (approved):
Tested on Monterey Bay on mac, and order looks good, ROMS runs.  Thanks

PR #643, opened 2026-07-23
Speed up nesting operations

ScottEilerman:
This PR modifies the internals of nesting-related grid operations for improved efficient. ~10x speedup is gained for both large and small grids.

- [ ] Closes #xxxx
- [ ] Tests added


kthyng (approved):
The parent crop mask might save enough time alone without changes to the interpolation. That said, I think this change is fine. I have often used delaunay triangulations in this sort of way. I also compared the old child nest grid and one resulting from this change and all of the variables were the same except for bathymetry, which is the same to order 1e-12:

Image

PR #653, opened 2026-08-08
JOSS Editorial Changes

jatkinson1000:
# Summary

Minor text and reference changes as part of the JOSS editorial process for https://github.com/openjournals/joss-reviews/issues/10234

One other suggestion is to consider writing Xarray and Dask.
This is how the two libraries refer to themselves inline in text. At the moment they are formatted as inline code, as would be `pip install`ed, so I have not made this change and leave it to you to decide. I think either option would be suitable in publication.

Corresponding author @NoraLoose


NoraLoose:
@jatkinson1000 thanks so much for these edits, they all look good to me.

> One other suggestion is to consider writing Xarray and Dask.

I will open a separate PR for this (and tag you there), because I don't think I can push to this PR.

jatkinson1000:
Super, no problem and, as I say, just an observation during reading - if you'd prefer to leave them as is I don't think it's a blocker so very much up to you.

PR #655, opened 2026-08-10
Update to docs and code for Unified BGC file v2.1

kthyng:
# Summary
Updated the Unified BGC file used in the docs to v2.1. This also required a few code changes to work with changes to the coordinates in the dataset and to allow use of the old version of the file as legacy.

## Breaking Changes
* The unified BGC dataset is now expected to be **v2.1 or later** (`BGCdataset_v2_1.nc`), which names its dimensions `longitude`/`latitude`/`depth` and stores `month` as an integer index 1-12. Earlier files are still read, but log a warning pointing to the [datasets documentation](https://roms-tools.readthedocs.io/en/latest/datasets.html#Downloading-the-Unified-BGC-Dataset). Because v2.1 no longer carries day-of-year values, `ROMS-Tools` places the twelve monthly records at the mid-month days already used for other monthly climatologies. Compared with output built from a pre-v2.1 file, this moves the source time axis by at most half a day, with three consequences: the `*_time` coordinates shift correspondingly; the seasonal dust deposition field, which is interpolated from the seasonal axis onto those monthly targets, shifts by up to ~0.4%; and anything interpolated *in time* from the climatology — `InitialConditions` at a given `ini_time`, for instance — picks up the slightly different interpolation weights. The twelve climatology records themselves are unchanged, so `BoundaryForcing` and `SurfaceForcing` tracer values are identical apart from their time labels.
* Note for users still on the oldest `BGCdataset.nc`: it predates the `temp_WOA`/`salt_WOA` fields, so it cannot be used for salinity restoring (`SurfaceForcing` with `type="restoring"`) or for `bgc_interpolation_method="density"`/`"density_mld"`.

## Bug Fixes
* The unified BGC dataset now actually receives the lateral dask chunking it asks for. The chunk sizes were keyed by the pre-v2.1 dimension names, and xarray silently ignores chunk keys that are not dimensions of the file, so a v2.1 file was read as a single lateral chunk. This is a correctness fix rather than a speed-up: a regional read is unaffected either way (the netCDF backend pushes the spatial selection down to the file, so the oversized chunk never materializes), but peak memory for operations that touch the full global field is roughly halved.

## Improvements
* `UnifiedDataset` reads both unified BGC file generations: dimensions are renamed only where a pre-v2.1 file still uses `lon`/`lat`/`dep`, and `month` is recognised as either a day-of-year or an integer month index. New `roms_tools.setup.utils.climatology_mid_month_days` helper gives monthly climatologies a single mid-month day convention, shared by `assign_dates_to_climatology` and the unified BGC dataset. A pre-v2.1 file's dimensions are swapped onto their coordinate variables rather than renamed, so `depth`/`latitude`/`longitude` come out indexed for either file generation (this also silences an xarray `UserWarning` on every legacy load).

## Miscellaneous
* Test data: `coarsened_UNIFIED_bgc_dataset_v2_1.nc` is added and used by the unified BGC fixtures; the pre-v2.1 coarse file stays registered so both layouts remain covered by the tests.
* Notebooks and dataset pages point at `BGCdataset_v2_1.nc`.

## Code Review Checklist


blsaenz (commented):
So 2.1 is pre-filled?  I remember talk of not pre-filling anything to keep as little invented data around (so no one mistakes it for real data).  I believe I also argued the other side of this, which was we wanted it to be fast.  I'm fine with whichever we are rolling with, as long as we note (in the NC file itself?) that there is filled data.

blsaenz:
Otherwise looks great.

kthyng:
@blsaenz I had done this with the "History" attribute in the file:
> This dataset provides initialization and boundary condition values for MARBL biogeochemical (BGC) tracers, aligned on the World Ocean Atlas 2023 (WOA23) 1 degree grid, as a 12-month climatology. Each variable's source data is annotated in the attrs. This dataset has been updated from v1 to include WOA23 nutrient values. Additionally, any nans in variables were filled using the xESMF inverse_dist method, and all variables were converted to float32.

But you reminded me to check on that and I realized that version was not what I uploaded to google drive! So thanks for the reminder.

kthyng:
I am going to merge because all I did was change the google drive url.

kthyng:
[fill_unified_bgc_file.ipynb](https://github.com/user-attachments/files/30947782/fill_unified_bgc_file.ipynb)

here is the notebook that was used to create v2.1 from v2

PR #656, opened 2026-08-11
Add date padding control via kwargs

blsaenz:
# Summary
This PR adds "start_time_pad" and "end_time_pad" KWARGs to time-dependent roms-tools setup classes that, when true, adds a single time increment before and after the start/end date of the class instance to support ROMS interpolation to those start/end dates (this is the current default behavior).  When a "time_pad" KWARG is False, no time padding takes place, and no data is generated beyond the corresponding start/end times of the class.  This allows users to make sequential calls to roms-tools setup classes with periods of exactly 1 month, or 1 year, and to write out those classes/periods without clobbering adjacent periods with time-padded files.  The default is True for both KWARGs, maintaining existing behavior.

## Breaking Changes
- N/A

## New Features
- KWARGs "start_time_pad" and "end_time_pad" were added to time-resolving roms-tools classes, defaulting to True. These KWARGs allows a user to switch off the existing behavior of adding time increments of data beyond designated start/end times, and to avoid potentially overwriting existing forcing files.  

## Bug Fixes

- N/A

## Improvements

- N/A

## Miscellaneous

- N/A

## Code Review Checklist


ScottEilerman:
The PR description says it adds `date_pad_start` and `date_pad_end` separately, but the actual code just uses a single `pad_times` to control both at once. Should it be as the description says? When looking at the code I was going to ask if there were going to be cases at the ends of a long sequence, where we do want the start pad for the first one in the set, and do want the end pad for the last one.

blsaenz:
@ScottEilerman You are too fast.  Not done yet!  I didn't even have time to label this a draft!

ScottEilerman:
Sorry! I saw it on the slack bot and figured I'd give it a quick look before getting sucked into other things.

kthyng:
@blsaenz It could be useful to be explicit that this fixes that issue with restart files or whatever motivates this. Also would it make sense to have some logic for when to use the pad kwarg or not? Or rather maybe the logic would be in the blueprint/workflow?

kthyng (approved):

PR #672, opened 2026-09-01
Multiple BGC sources

blsaenz:
# Summary

Supports multiple biogeochemical (BGC) sources for `BoundaryForcing` and `InitialConditions` roms-tools classes.  These changes should make it easier add, as well as mix-and-match, source data for feeding into ROMS BGC models.  This PR includes:
- Both `BoundaryForcing` and `InitialConditions` classes are essentially collections of `BoundaryForcingSource` and `InitialConditionsSource` classes, with the ability to inter-operate on them, for interpolation, etc.
- MARBL-specific information and processing is consolidated into a single class, which `BoundaryForcing` and `InitialConditions` leverage to generate output
- Support for import and passing of the BGCMarbl class that curates fields from multiple sources into MARBL tracers
- The wizard has separate panels for physics and BGC sources, where multiple BGC sources can be added or removed (For both `InitialConditions` and `BoundaryForcing`)
- Support for these BGC sources: UNIFIED, WOA, GLODAP, constant, ESPER, and CESM output files.
- ESPER support currently requires the [C]Worthy fork of PyESPER if be installed from source

## Breaking Changes

- **`BoundaryForcing` is now a multi-source wrapper; the previous class is `BoundaryForcingSource`.** The wrapper takes `source=` (physics) plus `bgc_sources=[…]` and exposes `.physics` / `.bgc` (a list). It has **no** `type=`, `physics_forcing=`, `.ds` or `.plot()`. Existing single-source code keeps working verbatim by renaming the class to `BoundaryForcingSource`.
- **`BoundaryForcing.save()` changed shape**: `save(physics_filepath, bgc_filepaths=None, …)` returning `(physics_paths, bgc_paths)`, rather than `save(filepath)`. Each BGC source is written to its own file (ROMS's `frcfiles` namelist key accepts a list), so there is no merge into one dataset as there is for initial conditions.
- **`bgc_model=` is now required whenever `bgc_source`/`bgc_sources` is given**, on both `InitialConditions` and `BoundaryForcing`; omitting it raises `ValueError`. Pass `bgc_model=rt.BGCMarbl` to reproduce previous behaviour.
- `InitialConditions` is a gentler change: it keeps `source`, `bgc_source`, `use_vars`, `.ds` and `.plot()`, and additionally gains `bgc_sources`, `bgc_model`, `.physics` and `.bgc`. Existing code needs only the `bgc_model=` addition above.
- New runtime dependency: **`threadpoolctl`**.

## New Features

- **Multiple BGC sources per object** via `bgc_sources=[{"source": {…}, "use_vars": […]}]`, each down-selectable with `use_vars` so sources can be combined without overlap.
- **`BGCModel` / `BGCMarbl`** — the BGC tracer set is now completed by a pluggable model class rather than hard-coded helpers, deriving every MARBL tracer from a single rule table.
- **WOA23 as a gridded BGC source** (`"WOA"`) for `InitialConditions` and `BoundaryForcing`, reading World Ocean Atlas 2023 nutrients and oxygen (`NO3`, `PO4`, `SiO3`, `O2`) directly from NCEI. WOA carries no carbon chemistry or iron, so it is meant to be combined with GLODAP, ESPER or a constants source for `DIC`/`ALK`/`Fe`. It also loads the matching monthly T/S to convert µmol kg⁻¹ → mmol m⁻³ and to supply the source density coordinate for `density`/`density_mld` interpolation; neither is written to ROMS output. Monthly WOA fields stop at 800 m (nutrients) and 1500 m (oxygen, T/S), so each variable is extended onto the full-depth 102-level annual grid per a new `deep_fill` option: `"annual_blend"` (default) splices the annual climatology underneath with a linear taper centred on each variable's own seam (half-width `deep_blend_halfwidth`, default 100 m), while `"ffill"` persists the deepest monthly value downward. Omitting `path` downloads to the `roms-tools` cache via the new `download_woa23_bgc()` helper.
- **ESPER BGC source** — derives BGC fields from the physics temperature/salinity already on the ROMS grid using PyESPER's empirical routines, with no dataset load and no lateral/vertical regridding (handled like the `constants` source). Configurable `method` (`lir`/`nn`/`mixed`) and `equation` (8/16). **Experimental**, and see the caveat below.
- **`serialize_dask` and `show_progress` on `save_datasets`**, and `serialize_dask` on the forcing objects' `save()` — a manual tool for writes whose per-chunk memory cost makes concurrent execution untenable.
- New public exports, all added to `docs/api.rst`: `BoundaryForcingSource`, `InitialConditionsSource`, `BGCModel`, `BGCMarbl`, and `compute_potential_density`.

## Bug Fixes

- **Regenerated a stale test.**
  `test_check_results[initial_conditions_with_bgc_from_climatology]` had been failing since the commit that last wrote it; the stored file's own metadata records that it was baked from a *dirty* working tree whose attribute set had since drifted, so it never matched any committed state. Values were never wrong — every difference was at most 1.17e-7 against a float32 eps of 1.19e-7 (one ulp), with 30 of 42 variables already bit-identical.

## Improvements

- **`group_by_month` groups by contiguous slice instead of `groupby`**, whose fancy indexing re-blocked the array. This removes the remaining redundancy: **2.8× → 1.0×**. Selection only — chunking is untouched.
- **`BGCMarbl` derives its tracers from one rule table** rather than two copies of the same math.
- **A missing or wrong PyESPER fails with actionable guidance, up front.** The check runs from `validate_esper_source` (called by `_input_checks`), so it is reported before the grid and physics regrid are built rather than minutes in, and it distinguishes "no PyESPER importable" from "importable, but it is upstream PyESPER rather than the fork that provides the `*_xr` methods" — the latter being the confusing case, since the package imports fine and only the attribute lookup fails.

## Miscellaneous

- Docs: `api.rst` gains the source classes and a Biogeochemistry section; `boundary_forcing.ipynb` and `initial_conditions.ipynb` gain a Quick Start section for the wrapper API; `end_to_end.ipynb`, `using_dask.ipynb` and `datasets_overview.rst` updated for the split.
- Test suite: **1127 passed, 55 skipped**.

### Reviewer caveats

- **ESPER requires CWorthy's PyESPER fork**, which is not on PyPI: only that fork ships the dask-native `lir_xr`/`nn_xr`/`mixed_xr` methods this adapter calls. It must be installed from a checkout (`pip install -e .`). Consequently **the ESPER estimation tests skip on CI** — they are gated behind `ROMS_TOOLS_PYESPER_PATH` pointing at a checkout containing `Mat_fullgrid/`. Without it the suite reports 1109 passed / 73 skipped; the 1127/55 figure above is with it set. Everything else, including the ESPER input-validation and chunk-plan tests, runs unconditionally.

## Code Review Checklist

- [ ] Closes #xxxx


roms_tools/datasets/download.py
```diff
@@ -303,3 +306,167 @@ def download_test_data(filename: str) -> str:
...
+        if key in ("temp_bgc", "salt_bgc"):
+            decade = ts_decade
+        for period in periods:
+            filename = woa23_filename(code, period, decade)
+            path = directory / filename
+            if path.exists() and not clobber:
+                continue
+            url = woa23_url(subdir, decade, filename)
+            logging.info("Downloading %s -> %s", url, path)
+            _download_one(url, path)
```
ScottEilerman, line 511: optional optimization: we could use multithreading or async to download many at once, if this is slow

roms_tools/datasets/lat_lon_datasets.py
```diff
@@ -1415,6 +1417,338 @@ def clean_up(self, ds: xr.Dataset) -> xr.Dataset:
...
+                "WOA temperature/salinity unavailable; using a uniform 1025 kg m-3 "
+                "for the umol/kg -> mmol/m3 conversion."
+            )
+            density = 1025.0
+
+        conversion_factor = density / 1000.0
+        for file_var in self.var_names.values():
+            if file_var in self.ds.data_vars:
+                self.ds[file_var] = self.ds[file_var] * conversion_factor
+
```
ScottEilerman, line 1739: noting that might claude had some questions about the above pieces and may have modified it -- flagging for us to look carefully at the final version for correctness

roms_tools/datasets/lat_lon_datasets.py
```diff
@@ -1930,6 +2264,267 @@ class UnifiedBGCDataset(UnifiedDataset):
...
+        Notes
+        -----
+        **Placeholder implementation** — returns a spatially uniform value of
+        1025 kg m⁻³ regardless of T/S.  Replace the body of this method with a
+        full equation-of-state (e.g. TEOS-10 via ``gsw``) when available.
+        The placeholder uses :func:`xarray.full_like` so the result is dask-lazy
+        and carries the same chunk structure as the inputs.
+        """
+        return xr.full_like(temperature, 1025.0)
```
ScottEilerman: double-checking whether this placeholder is fine for now or was intended to be replaced before merge?

roms_tools/datasets/lat_lon_datasets.py
```diff
@@ -1930,6 +2264,267 @@ class UnifiedBGCDataset(UnifiedDataset):
...
+                next(iter(self.ds.data_vars.values())),
+                next(iter(self.ds.data_vars.values())),
+            )
+
+        conversion_factor = density / 1000.0  # µmol kg⁻¹ → mmol m⁻³
+
+        for file_var in self.var_names.values():
+            if file_var in self.ds.data_vars:
+                self.ds[file_var] = self.ds[file_var] * conversion_factor
+
```
ScottEilerman, line 2524: same comment about sigma/density: let's make sure we double-check after the "fixes" PR that all of this stuff looks correct.

roms_tools/setup/bgc_model.py
```diff
@@ -0,0 +1,600 @@
...
+        "diazChl": 0.0075,  # mg m-3
+        "diazC": 0.0375,  # mmol m-3
+        "diazP": 0.01,  # mmol m-3
+        "diazFe": 7.5e-7,  # mmol m-3
+    }
+
+    # Fe → Lig multiplicative factor.
+    _FE_TO_LIG = 3.0
+
+    # Open-ocean background concentrations for tracers that have neither a source
```
ScottEilerman, line 260: is it worth a note for where these background concentrations came from?

roms_tools/datasets/lat_lon_datasets.py
ScottEilerman: looks like it has already been cleaned up in the fixes PR, nevermind.

ScottEilerman (commented):
review the code portion, made a few small comments, but no major blockers. i'll make a pr with the more careful review + fixes. i'd still like to take a look at the notebooks next week before we merge.

ScottEilerman (approved):

PR #675, opened 2026-09-05
Code review fixes

ScottEilerman:
### Summary

Follow-up to #672 from the review: fixes the two runtime blockers, makes the
multi-source contract explicit, switches tracer unit conversion to in-situ
density, and hardens the WOA/ESPER/download paths. No public API is renamed;
one return shape and one keyword removal are noted below.

### Breaking changes (relative to the #672 branch, not to any release)

- `BoundaryForcing.save()` returns `(physics_paths, bgc_paths)` where
  `bgc_paths` is `list[list[Path]]` of the files actually written (one list per
  source, in `.bgc` order). Previously it echoed the requested names, which did
  not exist when `group=True` split them.
- `BGCModel.process_bgc_fields(forcings)` no longer accepts `filepath=` /
  `serialize_dask=`. Saving is the wrapper's job.
- Two BGC sources supplying the same tracer (after derivation) is now a
  `ValueError`. Partition sources with `use_vars`. Previously IC silently took
  the first source and boundary forcing wrote the variable into both files.
- Tracer unit conversion for WOA23, GLODAP and ESPER uses in-situ density
  (`gsw.rho` at the pressure of each depth) rather than σ₀ + 1000. Deep values
  increase by up to ~1.7 % at 4000 m. The density *coordinate* used for
  `density`/`density_mld` interpolation is unchanged (still σ₀).

### Bug fixes

- `BoundaryForcing.save()` ignored `group`/`format` for BGC files (NETCDF4,
  monthly-grouped output regardless of the request).
- `serialize_dask=True` raised `ValueError` whenever `NUMBA_NUM_THREADS` was
  below `os.cpu_count()` (typical on pinned HPC allocations).
- Constant fill silently did nothing when the first companion had no tracer to
  use as a template, while logging that it had filled.
- `annual_blend` turned a valid monthly WOA value into NaN wherever the annual
  field was NaN (and vice versa).
- GLODAP without T/S files used a hard-coded 1025 through a placeholder with no
  warning.
- `roms_tools_git_commit` could record the *user's* repository (git walks up from
  site-packages in `.pixi`/`.venv` layouts) and wrote the string `'None'` into
  YAML when unknown.
- `_ensure_pyesper` left a stale entry on `sys.path` after a failed import.
- `test_esper_estimates_are_identical_under_time_chunking` never cut along time
  when the fork was installed (it patched a constant the code no longer reads).

### Improvements

- `BGCModel` is now an `abc.ABC`: `process_bgc_fields` is abstract (so `BGCModel()`
  raises at instantiation, not at first call), and `__init_subclass__` requires every
  concrete subclass to declare a non-empty `_TRACER_VARS` (an empty set would derive and
  fill nothing); intermediate abstract subclasses are exempt. Both wrappers validate
  `bgc_model=` **before** building the physics source via a new `validate_bgc_model`,
  with messages for the common mistakes (an instance instead of the class, an unrelated
  class, the abstract base). `_is_forcing` duck-typing is deliberately unchanged.
- The wrappers no longer hand-list the constructor arguments they forward to the
  physics source and the BGC companions (four near-identical dicts that had already
  diverged: `apply_2d_horizontal_fill` reached physics but not the companions). A new
  `forwardable_fields(wrapper_cls, source_cls, exclude=())` derives the set as the
  intersection of the two dataclasses' `init` fields; the only maintained input is the
  exclusion of same-name-different-meaning fields (`InitialConditions.use_vars`, the
  legacy single-source convenience). Snapshot tests pin the derived sets so a new field
  on either class forces a conscious decision; behavioural tests check a wrapper knob
  reaches physics and every companion.
- Unknown `bgc_sources` item keys (e.g. `use_var`) raise instead of being
  ignored and round-tripped into YAML.
- ESPER estimates only the tracers in `use_vars`, halving PyESPER work for
  carbon-only users and enlarging the per-chunk budget.
- Downloads: 60 s timeout, no retry on 4xx (with a pre-staging hint), a
  `Content-Length` check, guaranteed `.part` cleanup.
- `WOABGCDataset` is eager when `use_dask=False`, like every other dataset, and
  its file table is derived from `download.WOA23_BGC_VARIABLES`.
- Pre-5.0 YAML files fail with an explicit message (`BoundaryForcingSource`
  shape, or IC with `bgc_source(s)` and no `bgc_model`).
- Merged IC output records `bgc_sources` (names + `use_vars`) as a global attr.
- Dataclass `==` compares parameters, not datasets (`compare=False` on `ds`,
  `physics`, `bgc`, `ds_depth_coords`), so YAML round-trip asserts are real.
- `git rev-parse` runs once per process instead of twice per object.

### Tests

New: `TestBoundaryForcingWrapper` (save honours group/format, real paths,
length check, per-item interp override, YAML round trip); overlap raises for IC
and boundary layouts and for two `constants` sources; explicit Lig suppresses
derivation; unknown item key; first-template constant fill; thread clamp with
`NUMBA_NUM_THREADS` monkeypatched; in-situ vs potential density at 4000 m;
WOA conversion asserted at a non-zero depth; GLODAP fixture with T/S files and
an exact converted value; missing-T/S warning; blend NaN regression; seam value
pinned; ESPER `use_vars` narrowing (mocked); `_ensure_pyesper` rollback;
11 download tests without network; wrapper YAML round trip builds the physics source
exactly once (the wrapper file carries `bgc_sources` as raw item dicts and never the
companions' embedded `physics_forcing`, which only the standalone `*Source.to_yaml`
writes for solo reconstruction).

Full suite: **1155 passed, 86 skipped** without PyESPER (baseline on the branch:
1109 / 73). `pre-commit run --all-files` clean (ruff, mypy, absolufy).

### Adversarial review of this branch (medium effort, 5 findings)

- Fixed: `_ensure_pyesper` now also evicts the cached `PyESPER` module(s) from
  `sys.modules` on failure, so a corrected `source['path']` can retry in the same
  process.
- Fixed: `compute_in_situ_density` no longer broadcasts temp/salt to a common shape
  (only depth against lat); the only full-size allocation is the result, same footprint
  as the σ₀ conversion it replaced. Matters for the eager `use_dask=False` WOA path.
- Fixed: stale docstrings (`serialize_dask_and_boost_threads` still claimed a BLAS
  boost; `get_roms_tools_version_info` now states the once-per-process semantics).
- Accepted as-is: two lines of `docs/releases.md` were corrected. The repo rule is not
  to hand-edit that file, but #672 itself adds the whole 5.0.0 section by hand; the edits
  only fix wrong text inside that section (`constant` → `constants`, `save()` return
  shape). Ben may prefer to drop the section and rely on PR descriptions instead.
- One order-dependent new test (`test_bf_use_vars_narrows_…`) fixed by bypassing the NaN
  validation like its IC sibling.

### Left for the author / out of scope

- Whether the `annual_blend` taper should start *at* the seam rather than be
  centred on it (behaviour unchanged, now documented and pinned by a test).
- PyESPER fork distribution (not on PyPI); the 18 ESPER estimation tests still
  skip on CI.
- Notebook re-execution (`docs/*.ipynb` carry stale outputs).
- SODA's surface conversion is untouched (p = 0, identical result).


blsaenz (approved):
My own (necessary?) complications with bgc_model and other things in here makes me a bit sad, but I'm otherwise fine with the PR.  I wonder about our code needing things like this, but don't have any answers - what coding could replace the need all of this orchestration:

"The wrappers no longer hand-list the constructor arguments they forward to the
physics source and the BGC companions (four near-identical dicts that had already
diverged: apply_2d_horizontal_fill reached physics but not the companions). A new
forwardable_fields(wrapper_cls, source_cls, exclude=()) derives the set as the
intersection of the two dataclasses' init fields; the only maintained input is the
exclusion of same-name-different-meaning fields (InitialConditions.use_vars, the
legacy single-source convenience). Snapshot tests pin the derived sets so a new field
on either class forces a conscious decision; behavioural tests check a wrapper knob
reaches physics and every companion."

ScottEilerman:
> My own (necessary?) complications with bgc_model and other things in here makes me a bit sad, but I'm otherwise fine with the PR. I wonder about our code needing things like this, but don't have any answers - what coding could replace the need all of this orchestration:
> 
> "The wrappers no longer hand-list the constructor arguments they forward to the physics source and the BGC companions (four near-identical dicts that had already diverged: apply_2d_horizontal_fill reached physics but not the companions). A new forwardable_fields(wrapper_cls, source_cls, exclude=()) derives the set as the intersection of the two dataclasses' init fields; the only maintained input is the exclusion of same-name-different-meaning fields (InitialConditions.use_vars, the legacy single-source convenience). Snapshot tests pin the derived sets so a new field on either class forces a conscious decision; behavioural tests check a wrapper knob reaches physics and every companion."

I actually pushed Claude to make that change, rather than having ~20 lines each of hand-plumbed kwargs that purposefully excluded one or two that shouldn't be passed along... that becomes a recipe for failing to add a new kwarg in the right place and things silently getting dropped back to defaults. It's mind-numbing but it is what it is. I think once other priorities settle down, a more comprehensive design review / refactor could (not specifically of this work, but the repo as a whole) could clean up a lot of cruft.

PR #680, opened 2026-09-18
Fix zarr initial slice bounds

abigale-wyatt:
# Summary
River forcing for the PACMED domain was unable to be generated due to OOM errors. The cause was found to be the zarr branch of `_load_data_dask` (the only path used for the ARCO ERA5 cloud archive) which ignored `initial_slice_bounds` entirely, unlike the `open_mfdataset` branch. 

`RiverForcing` and `SurfaceForcing` both pass a bounding box here to keep ERA5 reads small, but it was silently dropped, so any ERA5 request loaded the full global (721x1440) grid regardless of the requested domain -- OOMing on `.compute()` for longer time ranges (confirmed: a 5-year river-temperature request tried to allocate a 170GiB array). This applies the same slicing the `open_mfdataset` branch already uses.

## Breaking Changes
- N/A

## New Features
- N/A

## Bug Fixes
- `_load_data_dask`'s zarr branch now applies `initial_slice_bounds` before returning, matching the `open_mfdataset` branch. Fixes OOMs when using `surface_forcing_source={"name": "ERA5"}` (`RiverForcing`) or `source={"name": "ERA5"}` (`SurfaceForcing`) for large domains or long time ranges against the ARCO cloud archive.
- `storage_options={"token": "anon"}` was passed unconditionally to `xr.open_zarr`, which the installed zarr/fsspec version rejects for any non-remote (local) path. Now only passed for `gs://`/`gcs://` sources.

## Improvements
- Removed a now-redundant eager `tair.chunk({"time": -1})` in `RiverForcing`'s ERA5 sampling (`_get_river_surface_temp`) -- `_smooth_and_floor_air_temp` already rechunks conditionally right before the one operation (a rolling mean) that needs a single time chunk.
- Deduplicated the `gs://`/`gcs://` remote-path check into a shared `_is_remote_gcs_path` helper in `utils.py`, used by both `_load_data_dask` and `resolve_era5_source`.

## Miscellaneous
- N/A

## Code Review Checklist


abigale-wyatt:
FYI- PACMED still gets OOM errors with a 5 year forcing likely because of its size.

blsaenz (approved):

abigale-wyatt:
It turns out that while its true that `initial_slice_bounds` silently does nothing for zarr loaded datasets, this doesn't help speed up anything with ERA because that dataset is natively pulling the whole Global grid on a single dask chunk per hour. No slicing is possible. This fix could be useful if we had other zarr sources, but doesn't help at all with ERA. Further, my attempt to run the PACMED `SurfaceForcing` using an explicit `initial_slice_bounds` with this fix results in an OOM error. Notably, the default mode which sets `initial_slice_bounds`  to `None` is unaffected. 

Either way, I suggest we just add info to the `ERA5ARCODataset` documentation that the source data cannot be sliced usefully and not merge this PR. It can be revisited if we ever develop another zarr data source.

